PR #8, opened 2017-11-02
Add example of component usage

prigara:


kotlin-react/README.md
```diff
@@ -64,4 +64,19 @@ class Welcome(props: WelcomeProps): RComponent<WelcomeProps, RState>() {
...
+
+And here's how we can use this component in another component:
+
+```kotlin
+import welcome.*
+
+class App : RComponent<RProps, RState>() {
+    override fun RBuilder.render() {
+        div("App-welcome") {
+            key="welcome"
```
Hypnosphi: You don't need keys when there's only one child (I actually have an idea how to make them unneeded for fragments as well)

And I think the usage example should be as simple as possible, like that:

```
fun RBuilder.app {
    welcome("Jane")
}
```

prigara:
@Hypnosphi done.

PR #16, opened 2017-11-20
Create Todo.kt

ScottHuangZL:
Add a simple Todo sample.

### DIFF
--- a/examples/src/main/kotlin/example/Todo.kt
+++ b/examples/src/main/kotlin/example/Todo.kt
@@ -28,6 +28,7 @@ class Todo(props: TodoProps) : RComponent<TodoProps, TodoState>(props) {
 
 
                 attrs {
+                    value = state.text
                     placeholder = "Input todo text here..."
                     onChangeFunction = {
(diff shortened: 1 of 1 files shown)


examples/src/main/kotlin/example/Todo.kt
```diff
@@ -0,0 +1,88 @@
+import kotlinx.html.*
+import kotlinx.html.js.*
+import org.w3c.dom.HTMLInputElement
+import react.*
+import react.dom.*
+
+interface TodoProps : RProps {
+    var items: MutableList<String?>
```
Hypnosphi: Let's call this prop `initialItems` (because it's only used for initialization), and use immutable `List` both here and in state. Then in `setState` you can add the items like that:
```
setState {
  items += text // returns a new List with `text` appended as last item
}
```

And remove them like that:
```
setState {
  items = items.filterIndexed { it != index }
}
```

ScottHuangZL: Sounds good. Thanks for your expertise support.

PR #23, opened 2017-11-25
Create AxiosSearch.kt

ScottHuangZL:
An example to show how to leverage **axios** lib to fetch remote data by Scott_Huang@qq.com (Zhiliang.Huang@gmail.com)
(Btw, Vue.js formally leverage axios for ajax related works, so it should be good if React.js also use it)
 

It will be helpful for people to use axios lib to make their static data web to live/real data web, thanks.

 Date: Nov 25, 2017

### DIFF
--- a/examples/src/main/kotlin/example/AxiosSearch.kt
+++ b/examples/src/main/kotlin/example/AxiosSearch.kt
@@ -35,6 +35,17 @@ external interface AxiosConfigSettings {
     var timeout: Number
 }
 
+external interface AxiosError {
+    var message: String
+}
+
+external interface AxiosResponse {
+    var data: dynamic
+    var status: Number
+    var statusText: String
+    var header: dynamic
+    var config: AxiosConfigSettings
+}
 
 class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props) {
     override fun AxiosState.init(props: AxiosProps) {
@@ -47,13 +58,13 @@ class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props)
             url = "http://ziptasticapi.com/" + zipCode
             timeout = 3000
         }
-        axios(config).then { response ->
+        axios(config).then { response: AxiosResponse ->
             setState {
                 zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
             }
-        }.catch { error: dynamic ->
+        }.catch { error: AxiosError ->
             setState {
-                zipResult = ZipResult("", "", "Find error, please open your console to learn detail.")
+                zipResult = ZipResult("Find error:", error.message, ". Please open your console to learn detail")
             }
             console.log(error)
         }
--- a/examples/src/main/kotlin/example/AxiosSearch.kt
+++ b/examples/src/main/kotlin/example/AxiosSearch.kt
@@ -61,12 +61,19 @@ external interface AxiosError {
     var message: String
 }
 
-external interface AxiosResponse {
-    var data: dynamic
-    var status: Number
-    var statusText: String
-    var headers: dynamic
-    var config: AxiosConfigSettings
+external interface AxiosResponse<T> {
+    val data: T
+    val status: Number
+    val statusText: String
+    val headers: dynamic
+    val config: AxiosConfigSettings
+}
+
+//just for this zip code search response data
+external interface Data {
+    val country: String
+    val state: String
+    val city: String
 }
 
 class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props) {
@@ -80,7 +87,7 @@ class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props)
             url = "http://ziptasticapi.com/" + zipCode
             timeout = 3000
         }
-        axios(config).then { response: AxiosResponse ->
+        axios(config).then { response: AxiosResponse<Data> ->
             setState {
                 zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
             }
--- a/examples/src/main/kotlin/example/AxiosSearch.kt
+++ b/examples/src/main/kotlin/example/AxiosSearch.kt
@@ -43,7 +43,7 @@ class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props)
     }
 
     private fun remoteSearchZip(zipCode: String) {
-        val config: AxiosConfigSettings = js {
+        val config: AxiosConfigSettings = jsObject {
             url = "http://ziptasticapi.com/" + zipCode
             timeout = 3000
         }
--- a/examples/src/main/kotlin/example/AxiosSearch.kt
+++ b/examples/src/main/kotlin/example/AxiosSearch.kt
@@ -1,0 +1,87 @@
+@file:Suppress("UnsafeCastFromDynamic")
+
+package example
+
+/**
+ * An example to show how to leverage axios lib to fetch remote data by Scott_Huang@qq.com (Zhiliang.Huang@gmail.com)
+ * (Btw, Vue.js formally leverage axios for ajax related works, so it should be good if React.js also use it)
+ *
+ * Date: Nov 25, 2017
+ */
+
+import kotlinext.js.js
+
+import kotlinx.html.*
+import kotlinx.html.js.*
+import org.w3c.dom.HTMLInputElement
+import react.*
+import react.dom.*
+
+data class ZipResult(val country: String, val state: String, val city: String)
+interface AxiosProps : RProps {
+}
+
+interface AxiosState : RState {
+    var zipCode: String
+    var zipResult: ZipResult
+}
+
+// You need input correct axios.min.js link address in public\static\index.html
+// I personally download axios.min.js proactive in put it into public\static\js folder
+// <script src="%PUBLIC_URL%/static/js/axios.min.js"></script>
+// or you can put   <script src="https://unpkg.com/axios/dist/axios.min.js"></script>     into index.html to avoid proactive download to local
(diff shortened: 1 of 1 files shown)


examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -0,0 +1,87 @@
+@file:Suppress("UnsafeCastFromDynamic")
+
+package example
+
+/**
+ * An example to show how to leverage axios lib to fetch remote data by Scott_Huang@qq.com (Zhiliang.Huang@gmail.com)
+ * (Btw, Vue.js formally leverage axios for ajax related works, so it should be good if React.js also use it)
```
Hypnosphi: React is a rendering library not a framework, so working with network is out of its scope

examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -0,0 +1,87 @@
...
+interface AxiosState : RState {
+    var zipCode: String
+    var zipResult: ZipResult
+}
+
+// You need input correct axios.min.js link address in public\static\index.html
+// I personally download axios.min.js proactive in put it into public\static\js folder
+// <script src="%PUBLIC_URL%/static/js/axios.min.js"></script>
+// or you can put   <script src="https://unpkg.com/axios/dist/axios.min.js"></script>     into index.html to avoid proactive download to local
+// And we provide a simple fun to wrap axios.js, it is not type safe, suggest JB team to provide a formal wrapping for this useful lib
```
Hypnosphi: Let's use commonjs instead:
```
@JsModule("axios")
```

examples/src/main/kotlin/example/AxiosSearch.kt
ScottHuangZL: Yeah, agree.
But better add axios example into kotlin-react for people to exchange remote data :)

examples/src/main/kotlin/example/AxiosSearch.kt
ScottHuangZL: Thanks for your advise!

examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -0,0 +1,84 @@
...
+        val infoText = "Input USA 5 digit zip code below... such as 90210 or 24124 or any valid zip code."
+        div {
+            p { +infoText }
+            input(type = InputType.text, name = "zipCode") {
+                key = "zipCode"
+                attrs {
+                    value = state.zipCode
+                    placeholder = "Input zip code ..."
+                    title = infoText
+                    onChangeFunction = {
```
Hypnosphi: Let's extract it into a component method

ScottHuangZL: Ok, please. You can direct change it. Thanks.

examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -0,0 +1,84 @@
...
+
+interface AxiosState : RState {
+    var zipCode: String
+    var zipResult: ZipResult
+}
+
+//Per Hypnosphi advice, change to common js way.
+//you should need "npm install axios --save" in advance in your project folder
+@JsModule("axios")
+external fun axios(config: dynamic): dynamic
```
Hypnosphi: We can add some basic typing right in the example:
```
external interface AxiosConfig {
  var url: String
  var timeout: Number
}

// T is the type of response data
external interface AxiosResponse<T> {
  data: T
}

@JsModule("axios")
 +external fun axios<T>(config: AxiosConfig): Promise<T>
```

examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -28,15 +27,48 @@ interface AxiosState : RState {
...
 
 class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props) {
     override fun AxiosState.init(props: AxiosProps) {
-        zipCode = "" 
+        zipCode = ""
         zipResult = ZipResult("", "", "")
     }
 
+    private fun remoteSearchZip(zipCode: String) {
+        val config: AxiosConfigSettings = js {
```
Hypnosphi: You can now use `jsObject`, which is type-safe

In fact, `js` is just a shortcut for `jsObject<dynamic>`

ScottHuangZL: Thanks for the great advice. Have change in code:)

Hypnosphi:
Would you like to add typings for response data as well?

ScottHuangZL:
Done to add typing both for axios response and error too.

After your review/comment, I get new skills:)

examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -47,13 +58,13 @@ class AxiosSearch(props: AxiosProps) : RComponent<AxiosProps, AxiosState>(props)
...
-        axios(config).then { response ->
+        axios(config).then { response: AxiosResponse ->
             setState {
                 zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
             }
-        }.catch { error: dynamic ->
+        }.catch { error: AxiosError ->
             setState {
-                zipResult = ZipResult("", "", "Find error, please open your console to learn detail.")
+                zipResult = ZipResult("Find error:", error.message, ". Please open your console to learn detail")
```
Hypnosphi: let's add an `errorMessage` field to state instead of misusing ZipResult data class

examples/src/main/kotlin/example/AxiosSearch.kt
```diff
@@ -35,6 +35,17 @@ external interface AxiosConfigSettings {
     var timeout: Number
 }
 
+external interface AxiosError {
+    var message: String
+}
+
+external interface AxiosResponse {
+    var data: dynamic
```
Hypnosphi: Let's make `axios` return `Promise<AxiosResponse>` instead of `dynamic.

We can also make both `AxiosResponse` and `axios` a generic type so that `data` has the type of type parameter not `dynamic`:

```
external interface AxiosResponse<T> {
  // we can use `val` instead of `var` here because we never assign those values ourselves
  val data: T
  ...
}

@JsModule("axios")
external fun <T> axios(config: AxiosConfigSettings): Promise<AxiosResponse<T>>
``` 

Then you can add type for data:
```
external interface Data {
  val country: String
  val state: String,
  val city: String
}
```

and use `axios` like this:
```
// the type for `response` should be inferred authomatically
axios<Data>(config).then { response -> ... }.catch { error: AxiosError -> ... }
```

ScottHuangZL: I don't think it is a good idea to add more typing due to below reason:

1. The axios not only get response, but also have catch error branch.
So, external fun <T> axios(config: AxiosConfigSettings): Promise<AxiosResponse<T>> not correct.

2. When I try hide/ignore catch branch firstly.  The compile show below error:
How to resolve it?
Can you show example code? Thanks.

src\App\AxiosSearch\AxiosSearch.kt:92:29: error: no value passed for parameter 'onFulfilled'
        axios<Data>(config).then { response ->
                            ^
src\App\AxiosSearch\AxiosSearch.kt:94:48: error: unresolved reference: data
                zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
                                               ^
src\App\AxiosSearch\AxiosSearch.kt:94:71: error: unresolved reference: data
                zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
                                                                       ^
src\App\AxiosSearch\AxiosSearch.kt:94:92: error: unresolved reference: data
                zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
                                                                                           ^
src\App\AxiosSearch\AxiosSearch.kt:96:34: error: unresolved reference: status
            console.log(response.status)
                                 ^
src\App\AxiosSearch\AxiosSearch.kt:97:34: error: unresolved reference: statusText
            console.log(response.statusText)

3. When I add back catch branch, it show below error:
src\App\AxiosSearch\AxiosSearch.kt:102:18: error: type inference failed: fun <S> catch(onRejected: (Throwable) -> S): Promise<S>
cannot be applied to
((AxiosError) -> Unit)

                .catch { error: AxiosError ->
                 ^
src\App\AxiosSearch\AxiosSearch.kt:102:24: error: type mismatch: inferred type is (AxiosError) -> Unit but (Throwable) -> Unit was expected
                .catch { error: AxiosError ->
                       ^
src\App\AxiosSearch\AxiosSearch.kt:102:26: error: expected parameter of type Throwable
                .catch { error: AxiosError ->

Hypnosphi: First, you have to put the `{ response -> ... }` lambda in parens, because it's not actually the last argument of `then`:

```
axios<Data>(config).then({ response: AxiosResponse<Data> ->
  setState {
    zipResult = ZipResult(response.data.country, response.data.state, response.data.city)
    errorMessage = ""
  }
})
```

Second, the `Promise` builtin type in Kotlin expects the failure argument to be a [`Throwable`](http://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-throwable/index.html). For this example it's OK, because we only use the `message` field:
```
.catch { error ->
  setState {
    zipResult = ZipResult("", "", "")
    // the Elvis operator is needed because `Throwable::message` is nullable
    errorMessage = error.message ?: ""
  }
}
```

ScottHuangZL:
Add generic type for Axios, it finally works!
Thanks for @Hypnosphi 's professional advice, does learn something from this practice:)

ScottHuangZL:
Thanks for approve the changes:)  Will will merge into branch? Thanks.

Hypnosphi:
Thanks @ScottHuangZL!

PR #38, opened 2018-02-19
Add redux wrapper

Jamdan2:
I just got started with the Store for now.
Dispatching actions, subscribing, and getting the state all work just fine.


kotlin-redux/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+val myStore = RStore(::myReducer, someState)
+```
+Access the state with the store's `state` field.
+
+```
+myStore.state
+```
+In Kotlin, Actions can be written as classes. Just implement the `RAction` interface.
```
Hypnosphi: Does that mean that they are non-serializeable?

kotlin-redux/README.md
```diff
@@ -0,0 +1,48 @@
...
+You can now dispatch the action as usual.
+
+```
+myStore.dispatch(MyAction())
+```
+
+Note that internally, kotlin-redux uses a single action type and reducer to dispatch the actions
+that you write in Kotlin. This is done to adapt Redux to a type-safe environment.
+
+To subscribe to your store, set the store's `listener` to a lambda expression. To unsubscribe, just
```
Hypnosphi: So you can only have one listener?

kotlin-redux/README.md
Jamdan2: To make the actions type-safe, it uses a single internal 'real' redux action to dispatch the actions that you write in kotlin.

kotlin-redux/README.md
Jamdan2: Good point. I will fix that

kotlin-redux/README.md
Hypnosphi: I wonder if there is any way to keep both typesafety and serializeability

kotlin-redux/README.md
Jamdan2: It works the same way as usual now.

kotlin-redux/src/main/kotlin/redux/Imports.kt
```diff
@@ -0,0 +1,9 @@
+@file:JsModule("redux")
+package redux
+
+@JsName("createStore")
+internal external class Store<out S>(reducer: (S, dynamic) -> S, preloadedState: S) {
```
Hypnosphi: Can it be this interface instead of `dynamic`?

```
{
  val type: String
  val action: RAction
}
```

Hypnosphi: `dispatch` should keep accepting `dynamic` (or maybe `Any`) though, because there are middleware

Jamdan2: It works for the most part, but the original dispatch() function expects a plain JavaScript object. Passing a kotlin object that extends an interface breaks it.

kotlin-redux/README.md
Jamdan2: Something like this is always an option:
```someAction = RAction(type = "sometype")```
but this would be ugly in a typesafe language like kotlin. If actions needed to be serialized, would it work to instance them like so?
```actionsToDispatch = listOf(SomeAction("someParameter"), ...)```
this would not capture the action in general, but would work for specific use cases of the action.

Comment (Hypnosphi) [reply kotlin-redux/README.md]: Sorry, I'm not sure I'm following

My question is — if we're using checks like `is MyAction` in reducer, how will it work with deserialized data?

File: kotlin-redux/src/main/kotlin/redux/Imports.kt (hunk 3 above)
Comment (Hypnosphi) [reply kotlin-redux/src/main/kotlin/redux/Imports.kt]: What about `reducer` signature?

File: kotlin-redux/README.md (hunk 1 above)
Comment (Jamdan2) [reply kotlin-redux/README.md]: Do you mean how would the parameters work?
```
data class AddToString(val string: String) : RAction

fun reducer(state: String, action: RAction) = when {
	is AddToString -> state + action.string // smart cast
	else -> state
}
```
is what I was thinking. I tested this, and it seems to work fine. I can change it though if it is a problem.

Comment (Hypnosphi) [reply kotlin-redux/README.md]: If `action` is deserialized from json, `is AddToString` check will fail

Comment (Hypnosphi) [reply kotlin-redux/README.md]: Related discussion:
https://github.com/reactjs/redux/issues/437

Comment (Jamdan2) [reply kotlin-redux/README.md]: Hmm... I tested it though, and it seems to work...
I think it's because we wrap it in an actual redux action, like so:
```
js {
	type = "WRAPPER_ACTION",
	action = SomeRAction()
}
```
It's a pure javascript object that has a more complex kotlin object as one of it's keys, that gets processed by the same actual redux reducer every time, which is the `wrapperReducer()` method that I wrote in the `RStore` class.

I realize that it would be preferable to use unique redux actions instead of a single one to execute all the fake redux actions that we write in kotlin, though...
I will try to think of a solution to this. Reflection, maybe?

Comment (Hypnosphi) [reply kotlin-redux/README.md]: How exactly did you test that?

Comment (Jamdan2) [reply kotlin-redux/README.md]: I made an example state, and example reducer, and some example actions, and then used the store to dispatch them, and got the results by logging it. No errors occurred, and the actions did what they intended.

Comment (Hypnosphi) [reply kotlin-redux/README.md]: So where's the serialization/deserialization part?

Comment (Jamdan2) [reply kotlin-redux/README.md]: The action class is instanced and stored under the `action` key in the wrapper action.

Comment (Hypnosphi) [reply kotlin-redux/README.md]: Serialization means means converting to string, and deserialization converting from string back to object. It's needed for features like replaying actions

Comment (Jamdan2) [reply kotlin-redux/README.md]: So basically add a toJson method?

Comment (Hypnosphi) [reply kotlin-redux/README.md]: Yes. And my concern is that after parsing the Json back, Kotlin type information will be lost

Comment (Jamdan2) [reply kotlin-redux/README.md]: It can be serialized into json via kotlinx.serialization. Also, to preserve type, a type key can be included in the json, which can then be checked for when converting the json back into a kotlin object.

Comment (Hypnosphi) [reply kotlin-redux/README.md]: how should usages of `deserializeAction` look like?

Comment (Jamdan2) [reply kotlin-redux/README.md]:
```
val jsonAction = "{\"type\":\"SomeAction\",\"someArg\":25}"
val action = deserializeAction<SomeAction>(jsonAction)
if (action != null) store.dispatch(action)
```

Comment (Hypnosphi) [reply kotlin-redux/README.md]: How do you know which type to use when deserializing?

Comment (Hypnosphi) [reply kotlin-redux/README.md]: My question is, how do I know that I should cast it to `SomeAction` not to `SomeOtherAction` if the string is only available at runtime

Comment (Jamdan2) [reply kotlin-redux/README.md]:
```
val action = when (JSON.parse((StringSerializer to StringSerializer).map, jsonAction)["type"]) = {
	"SomeAction" -> deserializeAction<SomeAction>(jsonAction)
	"SomeOtherAction" -> deserializeAction<SomeOtherAction>(jsonAction)
	else -> null
}
```
works, where action is a nullable RAction.

Comment (Hypnosphi) [reply kotlin-redux/README.md]: OK, please add it to readme

**Comment by Hypnosphi** [comment]
@bashor do you think it should be here or rather in https://github.com/Kotlin/js-externals ?

**Comment by Hypnosphi** [comment]
Maybe we could require actions to actually have a type field and use this instead of `"WRAPPER_ACTION"`?

Like this:
```
interface RAction {
  val type: String
}

fun dispatch(action: RAction) = store.dispatch(js {
  type = action.type
  this.action = action
})
```

This would improve redux devtools experience a lot. We could also advice users to use [sealed classes](http://kotlinlang.org/docs/reference/sealed-classes.html) for actions:

```
sealed class Action(override val type: String): RAction
class AddTodo(val text: String): Action("ADD_TODO")
class RemoveTodo(val id: Int): Action("REMOVE_TODO")
```

**Comment by bashor** [comment]
@Hypnosphi the current vision is that Kotlin/js-externals should contain only external declarations for the library with the minor exception for lightweight helpers usually inline functions.

**Comment by Jamdan2** [comment]
Wait, should I not include the serialization utilities themselves?

**Comment by Jamdan2** [comment]
@Hypnosphi I agree, WRAPPER_ACTION is not very nice. Although I think it would be nice if we could generate the type name by taking the class name and converting it to all caps with underscores, because that is what it would be anyway. What do you think?
```
fun dispatch(action: RAction) = store.dispatch(js {
  type = toActionName(action::class.simpleName) // for action DoStuff, would be DO_STUFF
  this.action = action
})
```
alternitavely, we could also just include the logic in the RAction interface.

**Comment by Hypnosphi** [comment]
> converting it to all caps with underscores, 

That's not actually necessary. If we're going to do generating, it's better to preserve names given by user as is

**Comment by Hypnosphi** [comment]
By the way, I think we should use a custom [store enhancer](https://github.com/reactjs/redux/blob/master/docs/Glossary.md#store-enhancer) to do the adjustment of dispatch/reducer API that RState class does currently

**Comment by Jamdan2** [comment]
Do you mean get rid of RStore altogether?

**Comment by Hypnosphi** [comment]
I mean, refactoring it from wrapper class to enhancer function

**Comment by Jamdan2** [comment]
Hmm... I'm not entirely sure what you mean. Can you show me what you're envisioning it would look like to create a store in this way?

File: kotlin-redux/src/main/kotlin/redux/Imports.kt [hunk 4]
```diff
@@ -0,0 +1,9 @@
+@file:JsModule("redux")
+package redux
+
+@JsName("createStore")
+internal external class Store<out S>(reducer: (S, WrapperAction) -> S, preloadedState: S) {
```
Hypnosphi: `createStore` is actually a plain function not a constructor function, so describing it as `class` doesn't seem correct

Hypnosphi:
Something like this

```kt
// typings mostly taken from https://github.com/reactjs/redux/blob/master/docs/Glossary.md

external interface Store<S, A> {
    fun getState(): S
    fun dispatch(action: A)
    fun subscribe(listener: () -> Unit): () -> Unit
}

val <S> Store<S, *>.state: S get() = getState()

typealias Reducer<S, A> = (S, A) -> S

typealias StoreCreator<S, A> = (Reducer<S, A>, S) -> Store<S, A>

fun <S> kotlinEnhancer(next: StoreCreator<S, WrapperAction>): StoreCreator<S, RAction> {
    return { reducer, initialState ->
        val wrapperReducer: Reducer<S, WrapperAction> = { state, action -> reducer(state, action.action) }
        val store = next(wrapperReducer, initialState)
        object: Store<S, RAction> {
            override fun getState() = store.getState()
            override fun dispatch(action: RAction) = store.dispatch(js {
                type = action::class.simpleName
                this.action = action
            } as WrapperAction)
            override fun subscribe(listener: () -> Unit) = store.subscribe(listener)
        }
    }
}
```


Then it should be used similarly to redux devTools](https://github.com/gaearon/redux-devtools/blob/master/docs/Walkthrough.md#storeconfigurestorejs)

```
val enhancer = compose(
  applyMiddleware(d1, d2, d3),
  ::kotlinEnhancer
)

createStore(rootReducer, initialState, enhancer)
```

Jamdan2:
I agree, that looks much better. It also allows for the user to use Redux traditionally if they so choose, which was something that I was trying to implement anyway. I'll modify it.

kotlin-redux/src/main/kotlin/redux/KotlinEnhancer.kt
```diff
@@ -0,0 +1,34 @@
...
+                @Suppress("UnsafeCastFromDynamic") // Needed because explicit cast causes illegal cast exception
+                store.dispatch(js {
+                    type = action::class.simpleName
+                    this.action = action
+                })
+                return action
+            }
+            override fun subscribe(listener: () -> Unit) = store.subscribe(listener)
+            override fun replaceReducer(nextReducer: Reducer<S, KotlinAction>) {
+                reducer = nextReducer
```
Hypnosphi: Original `replaceReducer` also dispatches an action: https://github.com/reactjs/redux/blob/master/src/createStore.js#L216

Let's reuse it like this:
```
override fun replaceReducer(nextReducer: Reducer<S, KotlinAction>) {
  store.replaceReducer(wrap(nextReducer))
}
```

kotlin-redux/src/main/kotlin/redux/KotlinEnhancer.kt
```diff
@@ -4,9 +4,7 @@ import kotlinext.js.js
 
 fun <S> kotlinEnhancer(next: StoreCreator<S, WrapperAction>): StoreCreator<S, KotlinAction> {
     return { reducer, preloadedState ->
-        @Suppress("NAME_SHADOWING") // Needed to allow replaceReducer to mutate reducer
-        var reducer = reducer
-        val wrapperReducer: Reducer<S, WrapperAction> = { state, action -> reducer(state, action.action) }
+        var wrapperReducer: Reducer<S, WrapperAction> = { state, action -> reducer(state, action.action) }
```
Hypnosphi: This is a nit, but I'd prefer extracting a top-level function to introducing a mutable variable:

```
val wrapReducer: (Reducer<S, KotlinAction>) -> Reducer<S, WrapperAction> { reducer ->
    { state, action -> reducer(state, action.action) }
}
```

Jamdan2:
Right now combine reducers does not work with kotlin actions, because there is no way to initialize the state. I think using middlewares to listen for the INIT action and then dispatching a KotlinAction version of INIT that can be used in kotlin reducers is the way to go to fix this.

Jamdan2:
I think this is good to go. Everything seems to be working nicely, including middleware. I think the next step would be to create a separate wrapper for react-redux.

Jamdan2:
@Hypnosphi Sorry having a bit of trouble with the build file, should be good now

Leonya:
Thanks, I have published it!

PR #70, opened 2018-07-26
Typesafe inline css

Jamdan2:
Adding this extension function allows for inline styles like in normal react, except typesafe. I have found it useful in one of my own projects. What do you think?

### DIFF
--- a/kotlin-react-dom/build.gradle
+++ b/kotlin-react-dom/build.gradle
@@ -4,5 +4,6 @@ configurePublishingWithNPM(react_dom_version)
 
 dependencies {
     compile project(":kotlin-react")
+    compile project(":kotlin-css:kotlin-css-js")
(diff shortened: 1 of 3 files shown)


Leonya:
@Hypnosphi are we OK with a dependency on kotlin-css in kotlin-react?

Hypnosphi:
Not really. How could we avoid it?

Leonya:
Well, we don't have to accept the PR as-is. We can simply add this helper function to the documentation. Alternatively, we could create a `kotlin-react-css-bridge` package, but that seems like too big a hassle for a utility function.

Jamdan2:
I agree, it is an ugly dependency. Do you want me to add the function to the README instead?

Leonya:
Yes, please.

PR #71, opened 2018-08-05
Redux build script problems

Jamdan2:
Hi, it appears that I forgot to apply kotlin2js to the gradle build of kotlin-redux, and the version of kotlinx serialization was not compatible with the current kotlin version. This was causing an empty jar to be published. It should work now.


Leonya:
I have bumped kotlinx.serialization to 0.6.1 (it was released a few hours ago).

Leonya:
Published, will you please test?

PR #75, opened 2018-08-16
Support for CSS3 hsl / hsla color declarations, plus useful functions

brettwooldridge:
Added support for CSS hsl/hsla color declarations.

Added functions:
 * ``lighten()``
 * ``darken()``
 * ``saturate()``
 * ``desaturate()``

Calculations are performed in HSL-space and converted back to RGB/RGBA as necessary.

I noticed the change to replace color names with hex values.  I already had this change on-deck to handle that (support color functions).

I feel it is essential that when a developer specifies a CSS color value of, for example, ``PaleVioletRed``, that ``PaleVioletRed`` is what should appear in the browser inspector.  I do a lot of work with themes, where there are dozens of colors, and it is invaluable to see the actual declared colors in the emitted CSS.

Along with all of the support for HSL, and convenience functions, this change does that.


brettwooldridge:
Forgot to mention, I also added support for rgb/rgba percentages versus 0-255 values.  The CSS specification allows for declarations like:
```
rgba(100%, 33%, 33%, 0.75)
```
The regex was updated to cover this case (without breaking existing semantics).

brettwooldridge:
@Leonya Do not merge yet.  I am adding unit tests, and have already uncovered one issue ... ``hsl`` values can be "unit-less", e.g. ``hsl(100, 100%, 100%)``, or the first parameter ([angle](https://developer.mozilla.org/en-US/docs/Web/CSS/angle)) can specify degrees (``deg``), gradians (``grad``), radians (``rad``), or turns (``turn``), which is currently not handled.

I'll update this pull requests when all unit tests are complete and passing.

kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt
```diff
@@ -89,237 +93,408 @@ enum class Clear {
...
-        val yellow = Color("#ffff00")
-        val yellowgreen = Color("#9acd32")
+        val currentColor = Color("currentColor")
+
+        // W3C predefined HTML colors (140), see the referenced specification above.
+        //
+        val aliceBlue = Color("AliceBlue", "#f0f8ff")
+        val antiqueWhite = Color("AntiqueWhite", "#faebd7")
+        val aqua = Color("Aqua", "#00ffff")
+        val aquaMarine = Color("AquaMarine", "#7fffd4")
```
Leonya: "aquamarine" is one word, it should not be capitalized as AquaMarine

kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt
```diff
@@ -89,237 +93,408 @@ enum class Clear {
...
+        val lightSkyBlue = Color("LightSkyBlue", "#87cefa")
+        val lightSlateGray = Color("LightSlateGray", "#778899")
+        val lightSteelBlue = Color("LightSteelBlue", "#b0c4de")
+        val lightYellow = Color("LightYellow", "#ffffe0")
+        val lime = Color("Lime", "#00ff00")
+        val limeGreen = Color("LimeGreen", "#32cd32")
+        val linen = Color("Linen", "#faf0e6")
+        val magenta = Color("Magenta", "#ff00ff")
+        val maroon = Color("Maroon", "#800000")
+        val mediumAquaMarine = Color("MediumAquaMarine", "#66cdaa")
```
Leonya: Here as well.

kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt
```diff
@@ -89,237 +93,408 @@ enum class Clear {
...
-        val turquoise = Color("#40e0d0")
-        val violet = Color("#ee82ee")
-        val wheat = Color("#f5deb3")
-        val white = Color("#ffffff")
-        val whitesmoke = Color("#f5f5f5")
-        val yellow = Color("#ffff00")
-        val yellowgreen = Color("#9acd32")
+        val currentColor = Color("currentColor")
+
+        // W3C predefined HTML colors (140), see the referenced specification above.
```
Leonya: I'm not sure about capitalization at all. The color names in the spec are in lowercase.

kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt
brettwooldridge: I was basing the color names off of [this table](http://www.html-color-names.com),  but [this table](https://www.htmlcsscolor.com/html-color-names) lower-cases ``Aquamarine``.

The W3C Color spec. lists all names completely lower-case (with no capital letters), but notes that they are indeed case-insensitive in practice.

I'm fine with either leaving it as is, changing it to ``Aquamarine``, or lower-casing everything.

brettwooldridge:
@Leonya I have updated the pull request, but have not addressed the case issues above.  The pull requests fixes many minor bugs and adds a fairly extensive set of unit tests.

kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt
Leonya: I see. Well, this table is not the official documentation. This one is: https://www.w3.org/TR/2018/REC-css-color-3-20180619/

Please note that there should be "grey" aliases for "gray" colors. So, 147 colors in total, not 140.

Leonya:
Great, and thanks for adding the test harness! Please fix the casing and I'll accept the PR.

brettwooldridge:
@Leonya By "fix the casing" do you want the names to be all lower-case, or just the changes to the specific names that you pointed out?

Leonya:
All lowercase.

brettwooldridge:
@Leonya Thanks!

kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt
```diff
@@ -89,237 +95,418 @@ enum class Clear {
...
 
-    fun withAlpha(alpha: Double): Color {
-        if (alpha < 0 || alpha > 1) {
-            throw IllegalArgumentException("Alpha should be a number between 0.0 (fully transparent) and 1.0 (fully opaque)")
-        } else {
-            val rgba = toRGBA()
-            return Color("rgba(${rgba.red},${rgba.green},${rgba.blue},${alpha * rgba.alpha})")
+    fun withAlpha(alpha: Double) =
+        when {
+            value.startsWith("hsl", true) -> with(fromHSLANotation()) { hsla(hue, saturation, lightness, normalizeAlpha(alpha) * this.alpha) }
```
Leonya: I don't get the `normalizeAlpha(alpha) * this.alpha` bit, I think it should be just `normalizeAlpha(alpha)`

brettwooldridge: The [original code](https://github.com/JetBrains/kotlin-wrappers/blob/50336a4fe3ad3251716616525071028cb1a635cf/kotlin-css/src/main/kotlin/kotlinx/css/StyleEnums.kt#L258) was:
```kotlin
return Color("rgba(${rgba.red},${rgba.green},${rgba.blue},${alpha * rgba.alpha})")
```
Bascially, multiplying the *new alpha* with the *original alpha*.  The ``normalizeAlpha()`` call just ensures that the new alpha is in the range of 0.0 to 1.0.

Leonya: Sorry, I get it now. I've added more tests and they all pass. Still, when I integrate the new build into my project, `blackAlpha` stops working correctly.

Leonya: This one proved to be quite tricky. The `normalize` function that was written using generic types was not working correctly in Kotlin/JS, after several attempts at fixing it I rewrote it without generics.

brettwooldridge: @Leonya Is that an issue that should be reported?

Leonya: Yes, I'll talk to the Kotlin team on Monday.

PR #78, opened 2018-08-21
kotlin-styled unit test

brettwooldridge:
@Leonya Do not commit, the unit test is failing, but I wanted your advice on how to test kotlin-styled...

### DIFF
--- a/kotlin-styled/src/test/kotlin/styled/TestNested.kt
+++ b/kotlin-styled/src/test/kotlin/styled/TestNested.kt
@@ -1,0 +1,34 @@
+package styled
+
+import kotlinext.js.asJsObject
+import kotlinx.css.Color
+import kotlinx.css.Display
+import react.RBuilder
+import react.RComponent
+import react.RProps
+import react.RState
+import kotlin.test.asserter
+
+
+class TestNested {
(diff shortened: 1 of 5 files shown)


kotlin-styled/src/test/kotlin/styled/TestNested.kt
```diff
@@ -0,0 +1,34 @@
...
+                            color = Color.black
+                        }
+                    }
+                }
+            }
+        }
+
+        val element = Element()
+
+        asserter.assertEquals("", "foo", element.asJsObject().toString())
```
brettwooldridge: @Leonya What goes here?   "foo" is obviously not the expected value, it's just a placeholder.  I'm not sure this is even the right approach (using an actual ``RComponent``).

At a minimum, the test does run and fail as part of the gradle build, which was a huge pain getting to work by itself.

aerialist7:
@Leonya, @brettwooldridge seems that this change is outdated.
@Leonya can you please close it?

aerialist7 (commented):
@Leonya, @brettwooldridge seems that this change is outdated.
@Leonya can you please close it?

PR #82, opened 2018-09-12
Allow css tag for everything (*).

Hamzantal:
This will help developers who need to style all tags.

### DIFF
--- a/kotlin-css/src/main/kotlin/kotlinx/css/TagSelectors.kt
+++ b/kotlin-css/src/main/kotlin/kotlinx/css/TagSelectors.kt
@@ -2,6 +2,8 @@ package kotlinx.css
 
 class TagSelector(val tagName: String)
 
+val star get() = TagSelector("*")
+
 val a get() = TagSelector("a")
 val abbr get() = TagSelector("abbr")
 val area get() = TagSelector("area")


Hypnosphi:
I wonder if we can utilize `*` operator overloading for that

kotlin-css/src/main/kotlin/kotlinx/css/TagSelectors.kt
```diff
@@ -2,6 +2,8 @@ package kotlinx.css
 
 class TagSelector(val tagName: String)
 
+val star get() = TagSelector("*")
```
Hypnosphi: It's not actually a tag

Hypnosphi: I think you should rather add following member function to https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-css/src/main/kotlin/kotlinx/css/CSSBuilder.kt#L34

```
fun star(block: RuleSet) = "*"(block)
```

The "star" name is also disputable. I would prefer something more semantic, like "everything" or "wildcard"

Leonya: Jeez, what was I thinking?

Hamzantal:
Oh my god. My bad. My apologies

PR #91, opened 2018-11-19
Rewrite gradle-files with Gradle Kotlin DSL

mkraynov:
I've rewritten gradle configuration files with Gradle Kotlin DSL.

Why? Because we are JetBrains, of course :) And because having full IDE support for gradle files is awesome.

All tasks are working as it was with Groovy, except that I can't check `bintrayUpload` and `npm_publish`, I have no rights and get 401 Unauthorized.

PR #98, opened 2019-01-17
Make example live

neworld:
I made this mainly because of a missing easy accessible playground to test bugfixes and improvements. Please, do not judge design, because there isn't any. Also, I copied the config from mine project and don't know how to access Gradle's `ext`. After #91 things could be easier.

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -1,6 +1,0 @@
-applyKotlinJS()
-
-dependencies {
-    compile project(':kotlin-react-dom')
-}
-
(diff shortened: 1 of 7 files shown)


steveme:
If the examples folder contained a number of example projects, i'd be happy to work through a fair few basic variations and populate with example projects using various combinations of features.

js-groovy-react
mpp-groovy-react
js-kts-react
mpp-kts-react
mpp-dce-kts-react-redux
mpp-kts-react-redux-hooks

etc. (or a better selection perhaps)

steveme:
Is there a way to get it to use the versions from the root folders versions.gradle?

aerialist7 (commented):
@Leonya, @neworld this change is outdated.
@Leonya can you please close it?

PR #120, opened 2019-07-06
Assign static class names to an element even if they are applied under `specific`

ezhaka:
For example, the following code

```
styledDiv {
    css {
        specific {
            +SomeStaticStyleSheet.someClass
        }
    }
}
```

produces something like `<div class="SomeStaticStyleSheet-someClass hFRAPv"></div>` from now on
instead of just `<div class="hFRAPv"></div>`.

It allows overriding some properties of the static class from the outside

```
object OverridingStyleSheet : StyleSheet("OverridingStyleSheet", isStatic = true) {
    val greenText by css {
    	descendants(SomeStaticStyleSheet.getClassSelector { it::someClass }) {
            color = Color.green
        }
    }
}
```

and not being afraid that `SomeStaticStyleSheet-someClass` won't be applied to the element because of nesting.

### DIFF
--- a/kotlin-css/src/main/kotlin/kotlinx/css/CSSBuilder.kt
+++ b/kotlin-css/src/main/kotlin/kotlinx/css/CSSBuilder.kt
@@ -4,14 +4,18 @@ typealias RuleSet = CSSBuilder.() -> Unit
 
 fun ruleSet(set: RuleSet) = set
 
-data class Rule(val selector: String, val block: RuleSet)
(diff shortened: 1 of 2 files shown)


ezhaka:
Warning: this PR may break existing library consumers as they could implicitly rely on the previous behavior.

ezhaka:
It works badly with pseudo-classes like `:active` and `:hover`: this fix makes the class under `:active` to be always on the element. The PR is not ready to be merged yet.

ezhaka:
Okay, it is only about `specific` now.

PR #121, opened 2019-07-22
React Hooks v16.8

Skolotsky:
Wrappers for react hooks and way to create functional component like this
```kotlin
val renderCount = functionalComponent<RProps> {
    val (count, setCount) = useState(0)
    useEffect {
        console.log("useEffect $count")
    }
    +"count $count"
    button {
        attrs {
            onClickFunction = {
                setCount(count + 1)
            }
        }
    }
}

fun RBuilder.renderCount() {
    child(renderCount)
}
```

### DIFF
--- a/kotlin-react/src/main/kotlin/react/Imports.kt
+++ b/kotlin-react/src/main/kotlin/react/Imports.kt
@@ -109,12 +109,6 @@ external fun rawUseLayoutEffect(effect: () -> dynamic, dependencies: RDependenci
 @JsName("useLayoutEffect")
 external fun rawUseLayoutEffect(effect: () -> dynamic)
 
-// Layout Effect Hook
-@JsName("useMutationEffect")
-external fun rawUseMutationEffect(effect: () -> dynamic, dependencies: RDependenciesArray)
-@JsName("useMutationEffect")
-external fun rawUseMutationEffect(effect: () -> dynamic)
-
 // Context Hook
 external fun <T> useContext(context: RContext<T>): T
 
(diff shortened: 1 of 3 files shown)


kotlin-react/src/main/kotlin/react/Imports.kt
```diff
@@ -86,3 +86,49 @@ external fun <P : RProps> lazy(loadComponent: () -> Promise<RClass<P>>): RClass<
...
+
+// State Hook
+@JsName("useState")
+external fun <T> rawUseState(initValue: T): RDependenciesArray
+
+// Reducer Hook
+@JsName("useReducer")
+external fun <S, A> rawUseReducer(reducer: RReducer<S, A>, initialState: S): RDependenciesArray
+@JsName("useReducer")
+external fun <S, A> rawUseReducer(reducer: RReducer<S, A>, initialState: S, initialAction: A): RDependenciesArray
```
ezhaka: Can we use `initialAction: A = definedExternally` and avoid duplicating the function?

Skolotsky: I don't sure that it's equal in converted js. And this "optimisation" is unnecessary.

Hypnosphi:
Hooks should only be defined inside a functional component body, so it's better to define them as extension functions

kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -120,3 +120,29 @@ fun <P : RProps> forwardRef(handler: RBuilder.(RProps, RRef) -> Unit): RClass<P>
         buildElements { handler(props, ref) }
     }
 }
+
+typealias FunctionalComponent<P> = (props: P) -> dynamic
+
+/**
+ * Get functional component from [func]
+ */
+fun <P : RProps> functionalComponent(
```
Hypnosphi: Does it replace `rFunction`?
https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-react/src/main/kotlin/react/ReactComponent.kt#L78-L85

PR #159, opened 2020-04-05
kotlin2js. Standard plugin configuration

turansky:
* Plugin in plugins DSL block
* Configuration via `plugins.withType`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,21 +41,21 @@ subprojects {
       "org.jetbrains.kotlin:kotlin-$module:$kotlin_version"
     }
 
-    ext.applyKotlinJS = {
-        apply plugin: 'kotlin2js'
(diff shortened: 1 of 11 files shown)


Leonya:
Please don't refactor build scripts, we are going to change them anyway when migrating to the new multiplatform plugin.

turansky:
> Please don't refactor build scripts
Do you mean `master` branch only?

Can I fix errors in `1.4-js-ir`?

Leonya:
> Do you mean master branch only?

Yes, these are outdated.

> Can I fix errors in 1.4-js-ir?

What errors do you mean?

turansky:
> What errors do you mean?
`mavenLocal()` usage for example
Fixed here #160

PR #183, opened 2020-04-13
Compile time setters (hide getters)

turansky:


kotlin-react-dom/src/main/kotlin/react/dom/RDOMBuilder.kt
```diff
@@ -105,13 +105,15 @@ open class RDOMBuilder<out T : Tag>(factory: (TagConsumer<Unit>) -> T) : RBuilde
     }
 
     var key: String
-        get() = props.key
+        @Deprecated(message = "Write only property", level = DeprecationLevel.HIDDEN)
```
Leonya, line 108: Please change to "Write-only"

turansky: Done

PR #187, opened 2020-04-14
Simplify 'hyphenize'

turansky:
[Playground](https://pl.kotl.in/vFU93tvmr)

### DIFF
--- a/kotlin-css/src/main/kotlin/kotlinx/css/Extensions.kt
+++ b/kotlin-css/src/main/kotlin/kotlinx/css/Extensions.kt
@@ -1,32 +1,10 @@
 package kotlinx.css
 
-private val hyphenize: (String) -> String = memoizeString {
-    buildString {
-        it.forEach {
-            append(
-                when (it) {
-                    in 'A'..'Z' -> "-${it.toLowerCase()}"
-                    else -> it
-                }
-            )
-        }
+private val CAPITAL_LETTER = Regex("[A-Z]")
+
+fun String.hyphenize(): String =
+    replace(CAPITAL_LETTER) {
+        "-${it.value.toLowerCase()}"
     }
-}
-
-private fun <R> memoizeString(fn: (String) -> R): (String) -> R {
-    val map = LinkedHashMap<String, R>()
-    return {
-        synchronized(map) {
-            if (!map.containsKey(it)) {
-                map[it] = fn(it)
-            }
-
-            map[it]!!
-        }
-    }
-}
-
-
-fun String.hyphenize() = hyphenize(this)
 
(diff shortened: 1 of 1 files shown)


kotlin-css/src/main/kotlin/kotlinx/css/Extensions.kt
```diff
@@ -1,32 +1,10 @@
...
-    return {
-        synchronized(map) {
-            if (!map.containsKey(it)) {
-                map[it] = fn(it)
-            }
+private val CAPITAL_LETTER = Regex("[A-Z]")
 
-            map[it]!!
-        }
+fun String.hyphenize(): String =
```
Leonya, line 5: @turansky do you mind writing a micro-benchmark comparing the old and the new versions? Yes, the new version is simpler, but I'm wondering if its performance is also OK. See, this function is called really often when constructing CSS, but the inputs are repeating all the time because the number of CSS properties is quite limited. Therefore memoization might be actually useful.

turansky: 1. Readability looks more important in this case
2. Previous code use `synchronized`

Leonya: > Readability looks more important in this case

Why? It's a simple private utility method.

> Previous code use synchronized

So?

turansky: [Simple test](https://pl.kotl.in/iJxZRL38D)

`7.000.000` calls
My results:
```
Before 2122 ms
After 1755 ms
```

Leonya: > Simple test

The link opens an empty playground.

turansky: > The link opens an empty playground.

It's a bug. Do they have issue tracker? :)
WA - Incognito mode

Leonya: OK, so it is significantly slower on the JVM after all, I was not completely wrong :)

Luckily, the regex version is faster in all browsers, so it's fine. Thanks for doing the tests!

> It's a bug. Do they have issue tracker? :)

I think it should be reported to the regular YouTrack project, KT.

turansky: First test - border case - 50% enum constants require correction

Popular case - [no correction required](https://pl.kotl.in/4WmuwDCGS)
```
// JS
Before 1632 ms
After 303 ms
```

PR #188, opened 2020-04-14
TagSelector factory

turansky:


kotlin-css/src/main/kotlin/kotlinx/css/TagSelectors.kt
```diff
@@ -1,116 +1,126 @@
...
+val textarea by TagSelectorFactory
+val tfoot by TagSelectorFactory
+val th by TagSelectorFactory
+val thead by TagSelectorFactory
+val time by TagSelectorFactory
+val tr by TagSelectorFactory
+
+val ul by TagSelectorFactory
 
 val varTag get() = TagSelector("var")
```
Leonya, line 125: I believe `val `var` by TagSelectorFactory` should work.

turansky: In that case 'var' usage won't be so comfortable :(

Leonya: It's a very rarely used HTML tag anyway, it's OK for it to be not so comfortable.

PR #200, opened 2020-04-15
Empty JS Object factory

turansky:


kotlin-styled/src/main/kotlin/styled/StyledComponents.kt
```diff
@@ -149,7 +149,7 @@ fun StyledComponents.injectGlobal(string: String) {
     val globalStyleComponent = createGlobalStyle(string)
     val element = window.document.body!!.appendChild(window.document.createElement("div")) as Element
     element.setAttribute("id", "sc-global-style-${globalStylesCounter++}")
-    val reactElement = createElement(globalStyleComponent, js {})
+    val reactElement = createElement(globalStyleComponent, jsObject())
```
Leonya, line 152: @turansky compilation is failing with 

```
> Task :kotlin-styled:compileKotlinJs FAILED
e: /Users/leonidkhachaturov/code/kotlin-wrappers/kotlin-styled/src/main/kotlin/styled/StyledComponents.kt: (152, 24): Type inference failed: Not enough information to infer parameter P in external fun <P : RProps> createElement(type: Any, props: P, vararg child: Any?): ReactElement
Please specify it explicitly.

e: /Users/leonidkhachaturov/code/kotlin-wrappers/kotlin-styled/src/main/kotlin/styled/StyledComponents.kt: (152, 60): Type inference failed: Not enough information to infer parameter T in inline fun <T : Any> jsObject(): T
Please specify it explicitly.
```

PR #221, opened 2020-04-18
redirect-router-dom: Make redirect.from optional

oxc:
react-router-dom RedirectComponent can also be used to render an
unconditional redirect. Make the from restriction optional.


Leonya:
Released in pre.98

PR #222, opened 2020-04-18
Merge duplicate Component.rClass properties

oxc:
There has been a KComponent.rClass property getter in React.kt for a while.
Recently a Component.rClass has been added (that thus also works for PureComponent), however it is not publicly available.
I don't see a reason why these should not be merged.

This change widenes the receiver parameters of the public rClass property, and removes the internal one.


Leonya:
Released in pre.98

PR #237, opened 2020-04-19
Sync JS Object declaration with specification

turansky:
[Specification](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object)


Leonya:
Released in pre.99

PR #256, opened 2020-04-21
Remove Children undocumented method

turansky:
Details - https://reactjs.org/docs/react-api.html#reactchildren


Leonya:
Released in pre.101

PR #260, opened 2020-04-21
Fix 'isStyledComponent' contract

turansky:
https://styled-components.com/docs/api#isstyledcomponent


Leonya:
Released in pre.101

PR #263, opened 2020-04-21
Delegate only for 'useState' hook

turansky:
```Kotlin
var count: Int by useState(13)
count = 42
println(count) // 42
```

### DIFF
--- a/kotlin-react/src/main/kotlin/react/hooks.kt
+++ b/kotlin-react/src/main/kotlin/react/hooks.kt
@@ -8,24 +8,17 @@ typealias RDependenciesList = List<dynamic>
 
 typealias RSetState<T> = (value: T) -> Unit
 
-fun <T> useState(initValue: T): Pair<T, RSetState<T>> {
-    val jsTuple = rawUseState(initValue)
-    val currentValue = jsTuple[0] as T
-    val setState = jsTuple[1] as RSetState<T>
-    return currentValue to setState
-}
+fun <T> useState(initValue: T): ReadWriteProperty<Any?, T> =
+    ReactStateDelegate(rawUseState(initValue))
 
-fun <T> useState(valueInitializer: () -> T): Pair<T, RSetState<T>> {
-    val jsTuple = rawUseState(valueInitializer)
-    val currentValue = jsTuple[0] as T
-    val setState = jsTuple[1] as RSetState<T>
-    return currentValue to setState
-}
+fun <T> useState(valueInitializer: () -> T): ReadWriteProperty<Any?, T> =
+    ReactStateDelegate(rawUseState(valueInitializer))
 
-private class ReactStateDelegate<T>(useState: Pair<T, RSetState<T>>) : ReadWriteProperty<Any?, T> {
-    private val state = useState.first
-
-    private val setState = useState.second
+private class ReactStateDelegate<T>(
+    stateTuple: RDependenciesArray
+) : ReadWriteProperty<Any?, T> {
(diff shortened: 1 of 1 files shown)


Skolotsky:
Why do you think that it should be only used with delegate?

Leonya:
I believe all state hooks should work the same. If this one only works as a delegate, then all of them should be delegates too.

Skolotsky:
@Leonya , can't agree with you - hooks are for different things. It doesn't make sense to make delegate for useEffect for example. But i think that we should preserve idiomatic react useState. And this PR isn't backward compatible anyway.

turansky:
> hooks are for different things

I agree, delegates could't be used for all hooks.

> Why do you think that it should be only used with delegate?

0. Classic Kotlin delegate use case?!
1. Delegate looks more useful in `useState` cases
```Kotlin
// Before
val (count, setCount) by useState(13)
setCount(42)
println(count) // 42

// After
var count: Int by useState(13)
count = 42
println(count) // 42
```
2. I don't see use cases where variant "Before" is better
3. If old variants required for 'smooth' migration, then I could add them with `_useState` name

Skolotsky:
There already is way to use delegate with `by state(...)`. We don't need to replace useState.

Leonya:
Right, sorry, they are not all the same.

I suggest closing this, it doesn't look like an improvement.

turansky:
@Skolotsky Do you have cases where current `useState` is better than delegate?

ilgonmic:
We can change name of delegate `state` on `useState`, maybe it will be more habitually for users
But I think, both delegate and pure function should be preserved.
Is is more user-friendly, and migration-curve will be more smoothly

ilgonmic:
And I agree that not all hooks can be used with delegates (for example `effect` or `reducer`)
But for consistency, I think that not only state hook should be delegated
For example `context`
```
val context by useContext(MyContext)
```

Skolotsky:
We shouldn't change behaviour of useState. It is already used and it will break compatibility without real reason. And it isn't obvious that useState isn't real equivalent of react "useState" but there is equivalent with another name. `val context by useContext(MyContext)` doesn't make sense bacause `val context = useContext(MyContext)` works fine, we don't need delegate here.

ilgonmic:
> We shouldn't change behaviour of useState
Agree, I don't debated about it

I told, that it can make sense to rename just __delegated__ function from `state` to `useState`, maybe through deprecation of `state` function
To do
```
val (a, setA) = useState(0) // valid
var b by useState(0) // valid too
var c by state(0) // don't know, maybe deprecated, maybe valid too
```

> val context by useContext(MyContext) doesn't make sense bacause val context = useContext(MyContext) works fine, we don't need delegate here.
It is just thoughts about consistency behavior for hooks

ilgonmic:
Ok, thanks @Skolotsky, in this case it will be declaration ambiguity with only return type difference

ilgonmic:
Although it can be done with extension
```
operator fun <T> Pair<T, RSetState<T>>.getValue(thisRef: Any?, property: KProperty<*>): T {
    return first
}
operator fun <T> Pair<T, RSetState<T>>.setValue(thisRef: Any?, property: KProperty<*>, value: T) {
        second(value)
}
```

It can be not beautiful, because it is extensions on all `Pair<T, (T) -> Unit>`, so maybe it will be useful in future to change `Pair` on own class with destructuring
for example
`data class RStateHook<T>(val value: T, val setValue: (T) -> Unit)`

Leonya:
Superseded by #264

PR #264, opened 2020-04-22
Add extensions to use useState as delegate

ilgonmic:
Additionally, deprecate state delegate
It leads to consistent usage of react hooks and it respect hooks naming with `use*`


turansky:
@ilgonmic 
```Kotlin
var count = useState(0)
println(count) // 0 or Pair? depends on import? dangerous!
```

Skolotsky:
As i see it doesn't depend on import. It always returns Pair, but there are setValue and getValue for that pair, so you can use pair as delegate with `by` keyword.

ilgonmic:
There is no ambiguity
It will be `Pair` because return value of `useState` preserved as `Pair`
There are extension operator methods, which add delegated functionality
```
var foo = useState(0)
println(foo) // Pair

var bar by useState(0)
println(bar) // 0
```

Leonya:
Released in pre.103

PR #265, opened 2020-04-22
Fix 'kotlinx-html' dependency usage

turansky:
Before:
* Dependency for all (9) subprojects

After:
* Dependency for `kotlin-react-dom` & `kotlin-styled`


Leonya:
Released in pre.103

PR #267, opened 2020-04-22
Remove unused repositories

turansky:


Leonya:
What about building with `1.4-Mn`? Isn't the eap repo needed for that?

turansky:
> What about building with `1.4-Mn`?
For `eap` testing this these repositories are required.

But `master` branch should't use these repositories, because `master` is **release** branch.

Leonya:
I get your point, but I don't think that having a perfect repository without a single line of "unnecessary" code is a reasonable goal.

PR #271, opened 2020-04-23
CI cron action

turansky:


Leonya:
Why build if there are no changes?

turansky:
There are external dependencies:
1. Gradle, NPM, Yarn (distributions)
2. Maven, NPM!! dependencies

Main reason - "Ready for release" check

Leonya:
But one can't release without first building locally anyway, so I still don't get the point. Seems like a waste of electricity. It's pretty normal for this repo not to receive any changes for several weeks. Once there's a change though, it will be built and released, and any build failures will be spotted and fixed anyway. I don't expect Gradle or React to disappear anytime soon, do you?

Issue #275, opened 2020-04-28
"Invalid hook call" occurred when add kotlin-styled to dependencies


yt8492:
I wrote functional component using hooks like this 

```kotlin
val component = functionalComponent<RProps> {
    val (inputText, setText) = useState("")
    textArea {
        attrs.onChangeFunction = {
            val target = it.target as HTMLTextAreaElement
            setText.invoke(target.value)
        }
    }
    div {
        + inputText
    }
}
```

This code works, but when I add `kotlin-styled` dependency, this code crashed.

Here is a sample code. You can check by comment-out `kotlin-styled` dependency and running `./gradlew browserRun`.
https://github.com/yt8492/kotlin-react-hooks-sample

pillimoj:
Please see #147 
It should be fixed by installing a versio of styled components lower than 5. e.g. `implementation(npm("styled-components", "4.4.0"))`

yt8492:
I see. Thank you!

yt8492:
1 more thing. That solution works at `org.jetbrains:kotlin-styled:1.0.0-pre.102-kotlin-1.3.72`, but version `1.0.0-pre.103-kotlin-1.3.72`, still occurred "Invalid hook call". I think this is another problem.

Leonya:
@turansky will you take a look please?

turansky:
Yes, I will.

turansky:
Dependency sync (`pre.103` for all) resolve the problem:
```
dependencies {
    implementation(kotlin("stdlib-js"))
    implementation("org.jetbrains:kotlin-react:16.13.0-pre.103-kotlin-1.3.72")
    implementation("org.jetbrains:kotlin-react-dom:16.13.0-pre.103-kotlin-1.3.72")
    implementation("org.jetbrains:kotlin-styled:1.0.0-pre.103-kotlin-1.3.72")
}
```

No more dependencies required for example. Compatible NPM dependencies already declared for wrappers.

yt8492:
@turansky Thank you. Your solution works on my first code, but when I use styled-component, "Uncaught Error" the same as #147 occurred.

```kotlin
val component = functionalComponent<RProps> {
    val (inputText, setText) = useState("")
    styledTextArea {
        css {
            borderWidth = 3.px
        }
        attrs.onChangeFunction = {
            val target = it.target as HTMLTextAreaElement
            setText.invoke(target.value)
        }
    }
    div {
        + inputText
    }
}
```
[...]

PR #280, opened 2020-04-30
Use compatible 'styled-components'

### LINKED ISSUE #147: Styled components css produces "Uncaught Error" with useState hook
When using styled-components and a useState hook, an `Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.` is produced when trying to set the state. 

Example code that should give error:
```kotlin
import kotlinx.css.Color
import kotlinx.css.backgroundColor
import kotlinx.html.js.onClickFunction
import react.RBuilder
import react.RProps
import react.child
import react.dom.button
import react.dom.p
import react.dom.render
import react.functionalComponent
import react.useState
import styled.css
import styled.styledDiv
import kotlin.browser.document

fun main() {
    render(document.getElementById("root")) {
        exampleStyledComponent()
    }
}

val exampleStyledComponent = functionalComponent<RProps> {
    val (count, setCount) = useState(1)

    styledDiv {
        css {
            backgroundColor = Color.red
        }
        button {
            attrs {
                onClickFunction = {
                    setCount(count + 1)
                }
            }
            +"Increment"
        }
        p { +"$count" }
    }
}

fun RBuilder.exampleStyledComponent() = child(exampleStyledComponent)
```

In this example, if I remove the `backgroundColor = Color.red` from the css lambda, `setCount` works.

I am using kotlin js 1.3.70-eap-274.

turansky:
Fix #147, #275

### DIFF
--- a/kotlin-styled/build.gradle.kts
+++ b/kotlin-styled/build.gradle.kts
@@ -13,5 +13,5 @@ dependencies {
 
     api(npm("css-in-js-utils", "^3.0.4"))
     api(npm("inline-style-prefixer", "^5.1.2"))
-    api(npm("styled-components", "^5.0.1"))
+    api(npm("styled-components", "^4.4.1"))
 }


Leonya:
Released in pre.104

PR #282, opened 2020-05-01
Add 'resize' css property

yt8492:
This pull request implements 'resize' css property.
https://developer.mozilla.org/en-US/docs/Web/CSS/resize


Leonya:
Released in pre.105

PR #283, opened 2020-05-06
Remove unused dependencies

turansky:


Leonya:
I have noticed that the tests under `kotlin-css/src/commonTest` are no longer executed when running `./gradlew test`. It used to work before the refactorings (I don't mean this particular commit but your earlier changes).

turansky:
Looks like 'feature' of new MPP plugin :)

```
./gradlew jsTest // works
./gradlew jvmTest // works
```

Leonya:
OK!

PR #285, opened 2020-05-06
Update README for kotlin-react

SebastianAigner:
Add installation instructions for Gradle-based projects
Add description of React entry point
Use functionalComponents instead of RBuilders in example code
Add usage instructions for function and class components using child function and two type of extensions
Add section on RBuilders for structuring complex components
Add section on React Hooks


Leonya:
Fantastic, thanks!

Issue #292, opened 2020-05-14
fontFace only renders once


IARI:
/commonMain/kotlinx/css/CSSBuilder.kt contains these:
```kotlin
// https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule
fun media(query: String, block: RuleSet) = "@media $query"(block)
fun supports(query: String, block: RuleSet) = "@supports $query"(block)
fun fontFace(query: String, block: RuleSet) = "@font-face $query"(block)
```
All of them in css are typically provided multiple times within the same context.
StringBuilder.buildRules however only creates a single CSSBuilder per 'selector'.
I guess with media and supports that works, because there's the 'query' parameter, which makes it a different selector.

Looking at the [css spec](https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#font-face-rule), we find the `@font-face` grammar rule:

> font_face_rule
>   : FONT_FACE_SYM S* '{' S* descriptor_declaration? [ ';' S* descriptor_declaration? ]* '}' S*
>   ;
> 

So for `@font-face` there is no query parameter allowed (the `S*` is just whitespace)
there should not be a `query` parameter in the api.

That aside, in its current state it seems to be unusable for the commmon use case of defining multiple fonts within se same root block.

If you want to define two @fontFace bodys, they all end up in the same body and values are just overwritten. @turansky this looks like a bug to me.

Leonya:
Yes, you're right. I'm not sure how to fix this.

IARI:
I can make a PR - you can look at it, and see if you see a problem with it.

Leonya:
Cool! Could you please write a test for it? `kotlin-css` has some.

IARI:
I fixed some stuff and added a test as requested - please have a look at it :)

PR #293, opened 2020-05-14
fix css fontFace - issue #292

IARI:
I suggest this fix for https://github.com/JetBrains/kotlin-wrappers/issues/292

Next to `rules: MutableList<Rule>` It introduces another List `multiRules: MutableList<Rule>` that is simply rendered after all the regular rules by `StringBuilder.buildRules`.

This solution does not ensure that all rules are rendered in the same order as they were specified.
I think this should in general not be a problem, except in a scenario like this:
1. the user specifies some rule `r1: Rule` as `multiRules` with a selector `s`
1. the user specifies a rule `r2: Rule` with the same selector `s` in `rules` after these
In regular css, `r1` would possibly be overwritten by `r2`, but here `r1` will instead appear later in the printed css code, and overwrite `r2`.
However, this scenario seems like a clear abuse of that system.

### DIFF
--- a/kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
+++ b/kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
@@ -43,9 +43,9 @@ interface RuleContainer {
     val rules: MutableList<Rule>
     val multiRules: MutableList<Rule>
 
-    fun rule(selector: String, block: RuleSet) = rule(selector, false, block)
+    fun rule(selector: String, block: RuleSet) = rule(selector, false, block = block)
 
-    fun rule(selector: String, passStaticClassesToParent: Boolean, block: RuleSet, repeatable: Boolean = false) =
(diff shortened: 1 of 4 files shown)


IARI:
Ok, apparently I forgot `KeyframesBuilder` which overrides `RuleContainer`, which i missed...
i need to fix this, it first, then i'll write a test.

kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
```diff
@@ -163,7 +177,7 @@ class CSSBuilder(
 
     fun supports(query: String, block: RuleSet) = "@supports $query"(block)
 
-    fun fontFace(query: String, block: RuleSet) = "@font-face $query"(block)
+    fun fontFace(block: RuleSet) = rule("@font-face", false, block, true)
```
Leonya, line 180: `block` should be the last parameter.

Leonya:
Released in pre.105

PR #299, opened 2020-05-21
useState. Avoid additional imports (getValue, setValue)

turansky:
```Kotlin
import react.getValue // redundant
import react.setValue // redundant
import react.useState

val (n, setN) = useState("n") // works

var k by useState("k") // works
```

### DIFF
--- a/kotlin-react/src/main/kotlin/react/hooks.kt
+++ b/kotlin-react/src/main/kotlin/react/hooks.kt
@@ -8,47 +8,37 @@ typealias RDependenciesList = List<dynamic>
 
 typealias RSetState<T> = (value: T) -> Unit
 
-fun <T> useState(initValue: T): Pair<T, RSetState<T>> {
+fun <T> useState(initValue: T): RStateDelegate<T> {
     val jsTuple = rawUseState(initValue)
-    val currentValue = jsTuple[0] as T
-    val setState = jsTuple[1] as RSetState<T>
-    return currentValue to setState
+    return RStateDelegate(
+        jsTuple[0] as T,
+        jsTuple[1] as RSetState<T>
+    )
 }
 
-fun <T> useState(valueInitializer: () -> T): Pair<T, RSetState<T>> {
+fun <T> useState(valueInitializer: () -> T): RStateDelegate<T> {
     val jsTuple = rawUseState(valueInitializer)
-    val currentValue = jsTuple[0] as T
-    val setState = jsTuple[1] as RSetState<T>
-    return currentValue to setState
+    return RStateDelegate(
+        jsTuple[0] as T,
+        jsTuple[1] as RSetState<T>
+    )
 }
 
-operator fun <T> Pair<T, RSetState<T>>.getValue(thisRef: Any?, property: KProperty<*>): T {
-    return first
-}
+class RStateDelegate<T>
+internal constructor(
+    private val state: T,
+    private val setState: RSetState<T>
+) : ReadWriteProperty<Any, T> {
+    operator fun component1(): T = state
+    operator fun component2(): RSetState<T> = setState
 
-operator fun <T> Pair<T, RSetState<T>>.setValue(thisRef: Any?, property: KProperty<*>, value: T) {
-    second(value)
-}
-
-private class ReactStateDelegate<T>(useState: Pair<T, RSetState<T>>) : ReadWriteProperty<Any?, T> {
-    private val state = useState.first
-    private val setState = useState.second
-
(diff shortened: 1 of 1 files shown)


kotlin-react/src/main/kotlin/react/hooks.kt
```diff
@@ -8,48 +8,38 @@ typealias RDependenciesList = List<dynamic>
...
+) : ReadWriteProperty<Any, T> {
+    operator fun component1(): T = state
+    operator fun component2(): RSetState<T> = setState
 
-private class ReactStateDelegate<T>(useState: Pair<T, RSetState<T>>) : ReadWriteProperty<Any?, T> {
-    private val state = useState.first
-    private val setState = useState.second
-
-    override operator fun getValue(thisRef: Any?, property: KProperty<*>): T =
+    override operator fun getValue(thisRef: Any, property: KProperty<*>): T =
```
ilgonmic, line 35: Are you sure that changing `Any?` on `Any` is valid?
Seems that it leads to error, because this delegate can be used in function scope without `thisRef`

turansky: Fixed

ilgonmic:
**Note**: It is breaking change, don't know when it should be merged
Maybe on react update, maybe on kotlin update (thanks to big 1.4 update :) ), but maybe even now, because I think minority uses exactly `Pair` without desctructuring.
Personally me, I think it can be done right now. Because on kotlin update users can faced with other build process problems, and changing of declaration can be additional frustration.

Skolotsky:
> **Note**: It is breaking change, don't know when it should be merged
> Maybe on react update, maybe on kotlin update (thanks to big 1.4 update :) ), but maybe even now, because I think minority uses exactly `Pair` without desctructuring.
> Personally I think it can be done right now. Because on kotlin update users can faced with other build process problems, and changing of declaration can be additional frustration.

I think right now is good moment.

Leonya:
Unfortunately, this code doesn't work in a real-world codebase.

```kotlin
inline fun <R> useLifetimed(crossinline builder: LifetimedHook.() -> R): R {
    js("debugger;")
    val onceHandlers by useState { mutableProperty<MutableList<Lifetimed.() -> Unit>?>(mutableListOf()) }

    return useLifetime(
        once = {
            onceHandlers.value?.forEach { it(this) }
            onceHandlers.value = null
        },
        eachTime = {
            val lifetime = lifetime
            val lifetimedEffect = object : LifetimedHook {
                override val lifetime: Lifetime = lifetime
                override fun once(handler: Lifetimed.() -> Unit) {
                    onceHandlers.value?.add(handler)
                }
            }
            lifetimedEffect.builder()
        }
    )
}
```

is compiled to

```javascript
  function extensionComponent$lambda($receiver, props) {
    var extension = useState(null);
    var onceHandlers = useState_0(useLifetimed$lambda);
    useLifetime(useLifetimed$lambda_0(onceHandlers), useLifetimed$lambda_1(onceHandlers, extensionComponent$lambda$lambda(props, extension)));
    track($receiver, void 0, void 0, extensionComponent$lambda$lambda_0(extension, props));
    return Unit;
  }

  function useLifetimed$lambda_0(closure$onceHandlers) {
    return function ($receiver) {
      var tmp$;
      if ((tmp$ = closure$onceHandlers.getValue_lrcp0p$(null, useLifetimed$onceHandlers_metadata).value) != null) {
        var tmp$_0;
        tmp$_0 = tmp$.iterator();
        while (tmp$_0.hasNext()) {
          var element = tmp$_0.next();
          element($receiver);
        }
      }closure$onceHandlers.getValue_lrcp0p$(null, useLifetimed$onceHandlers_metadata).value = null;
      return Unit;
    };
  }
```

and the error I'm getting is "Uncaught ReferenceError: useLifetimed$onceHandlers_metadata is not defined"

I'm guessing, this is a bug in the Kotlin compiler itself and I hope that @ilgonmic can will take a look. For now, I'm going to revert this and subsequent commits.

Leonya:
Reverted: https://github.com/JetBrains/kotlin-wrappers/commit/04dcc3f8394048f1152d010b581c2223c7a9ee45

turansky:
First suggestion - `inline` required when function returns non-stdlib delegate

ilgonmic:
Hmm, strange, as for me, I have `useLifetimed$onceHandlers_metadata` declared both in simple case, and seems that in your case too (tested without DCE)

```
var useLifetimed$onceHandlers_metadata = new PropertyMetadata('onceHandlers');
```

Anyway cc @bashor

PR #302, opened 2020-05-22
Sync parameter formatting

turansky:


Leonya:
There's no need to put each parameter on a separate line if they fit into line limit. The first change is OK (ProfilerProps), the rest aren't necessary.

PR #307, opened 2020-05-29
Support CDN delivery of kotlin-react-router-dom

### LINKED ISSUE #306: kotlin-react-router-dom doesn't support CDN delivery
In the [README.md of react-router-dom](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-dom#react-router-dom), I need to access window.ReactRouterDOM if I want to use a CDN. kotlin-react-router-dom doesn't support this usage.

subroh0508:
fix #306 

【example repository】
additonal methods/classes: https://github.com/subroh0508/colormaster/blob/externalize-react-router-dom/web/app/src/main/kotlin/utilities/imports_react_router_dom.kt

usage: https://github.com/subroh0508/colormaster/blob/85de3b97e1bad1ad01b12a1e7d1d3a9d0dc2c356/web/app/src/main/kotlin/containers/AppFrameContainer.kt#L27

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/imports.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/imports.kt
@@ -1,19 +1,34 @@
-@file:JsModule("react-router-dom")
-
 package react.router.dom
 
 import react.RClass
 import react.RProps
 
-external fun useHistory(): RouteResultHistory
+external interface ReactRouterDom {
+    @JsName("useHistory")
+    fun rawUseHistory(): RouteResultHistory
+    @JsName("useLocation")
+    fun rawUseLocation(): RouteResultLocation
+    @JsName("useParams")
+    fun rawUseParams(): dynamic
+    @JsName("useRouteMatch")
+    fun <T : RProps> rawUseRouteMatch(options: dynamic): RouteResultMatch<T>
+    @JsName("withRouter")
+    fun <T : RProps> rawWithRouter(component: dynamic): RClass<T>
+    @Suppress("PropertyName")
+    val HashRouter: RClass<RProps>
+    @Suppress("PropertyName")
+    val BrowserRouter: RClass<RProps>
+    @Suppress("PropertyName")
+    val Switch: RClass<RProps>
+    @Suppress("PropertyName")
+    val Route: RClass<dynamic>
+    @Suppress("PropertyName")
+    val Link: RClass<LinkProps>
+    @Suppress("PropertyName")
+    val NavLink: RClass<dynamic>
+    @Suppress("PropertyName")
+    val Redirect : RClass<RedirectProps>
+}
 
-external fun useLocation(): RouteResultLocation
-
-@JsName("useParams")
-external fun rawUseParams(): dynamic
-
-@JsName("useRouteMatch")
-external fun <T : RProps> rawUseRouteMatch(options: dynamic): RouteResultMatch<T>
-
-@JsName("withRouter")
-external fun <T : RProps> rawWithRouter(component: dynamic): RClass<T>
+@JsModule("react-router-dom")
+external val ReactRouterDomModule: ReactRouterDom
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/imports.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/imports.kt
@@ -5,8 +5,7 @@ package react.router.dom
 import react.RClass
 import react.RProps
 
-@JsModule("react-router-dom")
-external object ReactRouterDom {
+external interface ReactRouterDom {
     fun useHistory(): RouteResultHistory
     fun useLocation(): RouteResultLocation
     fun useParams(): dynamic
@@ -20,3 +19,6 @@ external object ReactRouterDom {
     val NavLink: RClass<dynamic>
     val Redirect : RClass<RedirectProps>
 }
+
+@JsModule("react-router-dom")
+external val ReactRouterDomModule: ReactRouterDom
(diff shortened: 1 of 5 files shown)


subroh0508:
@Skolotsky 

> Did you test it?

Yes! I implemented and tested the exact same method as the Pull Request in my repository(On `externalize-react-router-dom` branch) below.

https://github.com/subroh0508/colormaster/tree/externalize-react-router-dom

Both `browserDevelopmentWebpack` and `browserProductionWebpack` work fine.

kotlin-react-router-dom/src/main/kotlin/react/router/dom/imports.kt
```diff
@@ -1,19 +1,51 @@
...
 package react.router.dom
 
-import react.RClass
-import react.RProps
+import kotlinext.js.require
+import react.*
+import kotlin.browser.window
 
-external fun useHistory(): RouteResultHistory
+private val requireReactRouterDom get() = require("react-router-dom")
```
Skolotsky, line 9: It shouldn't work without additional webpack configuring.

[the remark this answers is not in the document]
subroh0508: Changed the implementation to not use `require("react-router-dom")`!

I also added some methods to pass the `ReactRouterDOM` retrieved from the window!

I'm sorry if my understanding of the points pointed out is wrong. Shouldn't that be a problem?

kotlin-react-router-dom/src/main/kotlin/react/router/dom/imports.kt
```diff
@@ -1,19 +1,24 @@
...
-@JsName("useParams")
-external fun rawUseParams(): dynamic
-
-@JsName("useRouteMatch")
-external fun <T : RProps> rawUseRouteMatch(options: dynamic): RouteResultMatch<T>
-
-@JsName("withRouter")
-external fun <T : RProps> rawWithRouter(component: dynamic): RClass<T>
+@JsModule("react-router-dom")
+external val ReactRouterDomModule: ReactRouterDom
```
ilgonmic, line 24: Do we really need it?
Could we use
```
@JsModule("react-router-dom")
external object ReactRouterDom {
...
}
```
Or do you see some caveats?

subroh0508: As you pointed out, Object was sufficient😇

ilgonmic:
@subroh0508 Could you please provide use-case for these changes?
Because for now I am unsure about it.
How do you plan to use it, and why `@JsModule` on file is inappropriate?

So in both cases for CDN, you need to declare necessary dependencies before application js file.
The once advantage which I can see is for condition
```
if (ReactRouterDom == undefined) {
  // which actions do you plan do here?
} else {
  // Here usual actions
}
```
But in case of CDN you declare `webpack.externals` and then you declare it in `index.html` before application script. And application script can find `window.ReactRouterDom`. Why do you need exactly condition?

ilgonmic:
@subroh0508 I didn't mean, that particular changes was redundant. Commit which you reverted, I think was useful, and it is exactly what I want from you.
But now I want to take a break and find cases, where it can be useful.
So please provide your use case, where you need this PR at all.
Why current support of CDN usage not enough for you?

Current support is:
1. Declare module as `webpack.externals`
0. Declare in `index.html`
 - `react`
 - `react-dom`
 - `react-router-dom`
0. Declare in `index.html`
 - Your `application.js`

So your `webpack.config.d` should look lik
```
config.externals = {
   React: "react",
   ReactDom: "react-dom",
   ReactRouterDom: "react-router-dom"
}
```

and index.html should look like
```
<script src="cdnprovider.org/react"></script>
<script src="cdnprovider.org/react-dom"></script>
<script src="cdnprovider.org/react-router-dom"></script>
<script src="cdnprovider.org/application"></script>
```

subroh0508:
@ilgonmic I'm sorry for the confusion.

> Could you please provide use-case for these changes?

The envisioned use-case is when the following description is added to the Webpack config.

```js
const WebpackCdnPlugin = require('webpack-cdn-plugin');
...
config.externals = [
  {
    'react-router-dom': 'ReactRouterDom',
  },
];
config.plugins.push(
  new WebpackCdnPlugin({
    modules: [
      {
        name: 'react-router-dom',
        var: 'ReactRouterDOM',
        path: 'umd/react-router-dom.min.js',
      },
    ],
    pathToNodeModules: path.resolve(rootPath, 'build/js'),
  }),
);
```

(full configs -> https://github.com/subroh0508/colormaster/blob/e02d88614c0b55f9291a30147069506d0e444ea1/web/app/webpack.config.d/01.cdn.js#L15-L52)

When I run the `browerRun` under this configuration file and launch the web app, I get a `TypeError` and it doesn't work properly. I wanted to fix this behavior, so I sent this PR.

```
Uncaught TypeError: Cannot read property 'withRouter' of undefined
    at eval (webpack-internal:///../../packages_imported/kotlin-react-router-dom/0.0.0/kotlin-react-router-dom.js:5)
    at eval (webpack-internal:///../../packages_imported/kotlin-react-router-dom/0.0.0/kotlin-react-router-dom.js:230)
    at Object.../../packages_imported/kotlin-react-router-dom/0.0.0/kotlin-react-router-dom.js (bundle.js:3162)
    at __webpack_require__ (bundle.js:30)
    at eval (webpack-internal:///./kotlin/colormaster-app.js:5424)
    at Object../kotlin/colormaster-app.js (bundle.js:3404)
    at __webpack_require__ (bundle.js:30)
    at Object.0 (bundle.js:3416)
    at __webpack_require__ (bundle.js:30)
    at bundle.js:94
```

subroh0508:
**However**, when I checked the operation with `pre.104-kotlin-1.3.72`, I was able to confirm the correct operation with config.externals enabled without modifying the code.

It turns out that the cause of the `TypeError` was the typo in my webpack config😭

I am so sorry for putting you through my simple mistake🥺🥺🥺

subroh0508:
@ilgonmic 
> So as I understand it is good enough, and this PR is not required for your use case?

Yes. All the problems I had were solved with the existing code. Thank you for the review!

ilgonmic:
Thank you for participating!

PR #309, opened 2020-06-06
Support setter of BrowserRouter / HashRouter attributes

subroh0508:
Enabled setting the attribute values of BrowserRouter and HashRouter.

ref. https://reacttraining.com/react-router/web/api/BrowserRouter
ref. https://reacttraining.com/react-router/web/api/HashRouter


Leonya:
Released in pre.109

PR #310, opened 2020-06-13
'toString()' instead of 'value'

turansky:


Leonya:
Thanks!

Leonya:
Released in pre.109

PR #311, opened 2020-06-13
styled-components. Prepare IR support

turansky:
Additional:
* Use standard name for default import (like in docs)


Leonya:
This broke compilation.

turansky:
Compilation fixed

Leonya:
Released in pre.109

PR #313, opened 2020-06-20
Fix styled method's argument RClass -> Any

subroh0508:
**AsIs**

- Only `RClass` can pass `styled` method

**ToBe**

- Allows we to pass other classes to `styled` method(like `FunctionalComponent`)


ilgonmic:
Is there problem only with `FunctionalComponent`?
Why not use polymorphism, 2 functions each get parameter with one of `RClass<P>`, another with `FunctionalComponent<P>`?

subroh0508:
@ilgonmic Because I can pass `Any` type arguments to the constructor of `StyledElementBuilder`, I implemented  `styled` with `Any` instead of polymorphism.

If it's more appropriate to use polymorphism, I'll fix it!

PR #319, opened 2020-07-04
Reducing functional component verbosity

andylamax:


kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -47,6 +47,11 @@ open class RBuilder {
     ): ReactElement =
         child(this, jsObject(), handler)
 
+    operator fun <P : RProps> FunctionalComponent<P>.invoke(
```
Skolotsky, line 50: Invoking functional component like this will not check if you pass all required props or not. So there will be lots of places where can be mistake. Writing wrapper function  make one point where you need function parameters similar to properties structure and if it is right - it will be right in all places where it called.

buildSrc/local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
+# as it contains information specific to your local configuration.
+#
+# Location of the SDK. This is only used by Gradle.
+# For customization when using a Version Control System, please read the
+# header note.
+#Sun Jul 05 00:12:06 EAT 2020
+sdk.dir=/media/andylamax/63C23C360914D355/Installers/cross-platform/Sdk
```
Skolotsky: It shouldn't be in PR

kotlin-react/src/main/kotlin/react/RBuilder.kt
andylamax: How do you suggest we go about it to reduce the verbosity of the function? Coz, using hooks in kotlin-react like this is so verbose.

Skolotsky: You need to have now wrapper for calling component. It's only way i see.

```kotlin
val dumbComp = functionalComponent<DumbProps>{ props->
   val count by useState(0) // hooks and effects here
}

fun RBuilder.DumbComp(name: String) = child(dumbComp){ // so dumbComp will be the same each time
  attrs {
    this.name = name
  }
}
```
I don't see any way to reduce verbosity without compiler plugin... But we haven't it for now.

PR #340, opened 2020-08-26
Add hex function for Color with unit test

Wernerson:
I added a `hex` color function like `rgb` and `hsl` for Kotlin styled.
As parameter it takes an integer.
It is meant to be used like `hex(0xdeadbe)`.
I think it's a useful function as I use a lot of hex codes and using a function is more idiomatic than `Color("#deadbe")`.


Leonya (approved):

Leonya:
Released in pre.113

PR #341, opened 2020-08-27
Do not inline react.rFunction builder

bnorm:
By inlining `rFunction`, React Developer Tools Chrome extension shows any hooks deeply nested within some structure. Not inlining removes these extra layers of structure, which make debugging component hooks much easier.

Here's what it looks like before:

image

Here's what it looks like after:

image

This is how `functionalComponent` works (doesn't inline) but that builder doesn't support naming the component which gives a tree of `Anonymous` components which is worse. The other option would be to add `displayName` support to `functionalComponent` if that is the direction this library is moving and the goal is to remove `rFunction`.


Skolotsky (approved):

Leonya:
Released in pre.113

PR #343, opened 2020-09-10
Add ios target and workaround native plugin late evaluation

ptmt:
This 3 explicit targets are required for now so that kotlin-css can be consumed by Kotlin/Native ios targets. Can be extended later for all native targets, if needed.


Leonya (approved):

Leonya:
Released in pre.114

PR #345, opened 2020-09-17
Update to styled components v5

Skolotsky:
There were two problems:
1) Animation keyframes internal rules generation changed
2) `createGlobalStyle` has internal check of using in render by hook `useRef` and it was called conditionally against hooks rules.

And just added ability to name functionalComponent

### DIFF
--- a/kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
+++ b/kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
@@ -169,7 +169,7 @@ class CSSBuilder(
     }
 
     fun prefix(selector: String, block: RuleSet) {
-        "$selector &"(block)
+        "$selector &&"(block)
     }
 
     // https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule
--- a/kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
+++ b/kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
@@ -169,6 +169,7 @@ class CSSBuilder(
     }
 
     fun prefix(selector: String, block: RuleSet) {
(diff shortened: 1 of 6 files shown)


kotlin-css/src/commonMain/kotlin/kotlinx/css/CSSBuilder.kt
```diff
@@ -169,7 +169,7 @@ class CSSBuilder(
     }
 
     fun prefix(selector: String, block: RuleSet) {
-        "$selector &"(block)
+        "$selector &&"(block)
```
Leonya, line 172:
```suggestion
        // Temporarily using && here because of a bug introduced in version 5.2: https://github.com/styled-components/styled-components/issues/3244#issuecomment-687676703
        "$selector &&"(block)
```

kotlin-styled/src/main/kotlin/styled/StyledComponents.kt
```diff
@@ -81,29 +81,70 @@ inline fun <P : CustomStyledProps> RElementBuilder<P>.css(noinline handler: Rule
...
-private var globalStylesCounter = 0
+private object GlobalStyles {
+    private val component = functionalComponent<GlobalStylesComponentProps> { props ->
+        props.globalStyles.forEach {
+            child(it, jsObject {}, emptyList())
+        }
+    }
+
+    private val root by kotlin.lazy {
+        val element = window.document.body!!.appendChild(window.document.createElement("div")) as Element
```
Leonya, line 119:
```suggestion
        val element = document.body!!.appendChild(document.createElement("div")) as Element
```

Leonya (approved):

kotlin-styled/src/main/kotlin/styled/StyledComponents.kt
Skolotsky: Check showed that `document` can't be resolved

Leonya (approved):

Leonya:
Released in pre.116

PR #348, opened 2020-09-25
Remove reflection hardcode from kotlin-styled for compatibility with both compiler backends

ilgonmic:
Now save `holders` and `_properties` in memory

Open question: 
- There are no removes for this mutable collections `holders` and `_properties`, do you think if removals are required, and does redundancy of removals operations lead to memory leaks?


Leonya (approved):

Leonya:
Yes, I believe cleaning up after injecting is a good idea.

Leonya:
Released in pre.121

PR #353, opened 2020-10-12
Fix inline styles to work with JS_IR

anton-bannykh:
The JS_IR compiler tends to rename locals more often
compared to the old one. This change makes the code
work even if `value` gets renamed to `value_N`.


Leonya (approved):

Leonya:
Released in pre.124

PR #378, opened 2020-12-20
Move `Record` to base classes

turansky:
Original `Record` - https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype

\+ @vlsi


Leonya:
Released in pre.133

PR #391, opened 2021-01-14
`useMemo` with "kotlin" signature

turansky:


Skolotsky:
Change is fine, but it looks like wee need such signature for another hooks and cases. For example edge case where there aren't dependencies (not empty array, but undefined instead of it).

turansky:
```
// 1. called once
useMemo(callback, [])

// 2. called every render 
useMemo(callback)
```

@Skolotsky 
1. Do you use `useMemo` with `undefined` dependencies?
2. Which variant is more used `[]` or `undefined`?

Skolotsky:
Still what about useEffect, useCallback? Can you bring same signature for them too?

turansky:
> Still what about useEffect, useCallback? Can you bring same signature for them too?

`useCallback` - yes
`useEffect` - yes, but it will displace [current methods](https://github.com/JetBrains/kotlin-wrappers/blob/ebb9fca25d57d88dc856426d8dcebe70f3b1f750/kotlin-react/src/main/kotlin/react/hooks.kt#L66) (breaking change). Will you approve it?

Skolotsky:
Why `useMemo` doesn't displace them, but `useEffect` will displace? They have nearly same signature with `useMemo` now. So there will be two types of signature (which isn't very good too).

turansky:
> Why `useMemo` doesn't displace them, but `useEffect` will displace? 

Because `useEffect` have 2 existed "adapters" and only raw declaration for `useMemo`

Skolotsky:
Don't understand why there can't be another `useEffect` signature and another `useEffectWithCleanup` signature? There shouldn't be ambiguity at call site.

turansky:
Could we deprecate old `useEffect` and `useEffectWithCleanup` adapters in this case?

Skolotsky:
I think we can deprecate them in case when it will be possible to cover case with `undefined` dependencies with new signature. For `useLayoutEffect` it can be useful to run effect after each render.

turansky:
@Skolotsky I have no access to PR's fork (new fork create)
Could we start from this partial change?

Next changes I will create in separate reviews.

Skolotsky (approved):

Leonya:
Released in pre.141

PR #392, opened 2021-01-15
CSS. Add `appearance` support

turansky:
[appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance)


Leonya (approved):

Leonya:
Released in pre.139

PR #399, opened 2021-01-25
Add `ReadonlyArray` typealias

turansky:
[Name source](https://www.typescriptlang.org/docs/handbook/2/objects.html#the-readonlyarray-type)


Leonya:
Released in pre.142

Issue #366, opened 2020-11-08
RMutableRef should extend RReadableRef, not RRef


fluidsonic:
```kotlin
external interface RMutableRef<T> : RReadableRef<T> {
    override var current: T
}
```

PR #412, opened 2021-02-15
#366 Fix ref hierarchy

turansky:
Fix for #366

### DIFF
--- a/kotlin-react/src/main/kotlin/react/Imports.kt
+++ b/kotlin-react/src/main/kotlin/react/Imports.kt
@@ -68,7 +68,7 @@ external val Fragment: RClass<RProps>
 external fun <T> createContext(defaultValue: T = definedExternally): RContext<T>
 
 // Refs (16.3+)
-external fun <T> createRef(): RReadableRef<T>
+external fun <T: Any> createRef(): RReadableRef<T>
 
 @JsName("forwardRef")
 external fun <P : RProps> rawForwardRef(forward: (props: P, ref: RRef) -> Any): RClass<P>
@@ -139,11 +139,11 @@ external fun <T : Function<*>> useCallback(callback: T, dependencies: RDependenc
 external fun <T> useMemo(callback: () -> T, dependencies: RDependenciesArray): T
 
 // Ref Hook (16.8+)
-external interface RMutableRef<T> : RRef {
-    var current: T
+external interface RMutableRef<T: Any> : RReadableRef<T> {
+    override var current: T?
 }
 
-external fun <T> useRef(initialValue: T): RMutableRef<T>
+external fun <T: Any> useRef(initialValue: T): RMutableRef<T>
 
 // Imperative Methods Hook (16.8+)
 external fun useImperativeHandle(ref: RRef, createInstance: () -> dynamic, inputs: RDependenciesArray)
--- a/kotlin-react/src/main/kotlin/react/ReactComponent.kt
+++ b/kotlin-react/src/main/kotlin/react/ReactComponent.kt
@@ -74,7 +74,7 @@ external interface RContext<T> {
 // Refs (16.3+)
 external interface RRef
 
-external interface RReadableRef<out T> : RRef {
+external interface RReadableRef<out T: Any> : RRef {
     val current: T?
 }
 


Skolotsky (approved):

Skolotsky:
Not sure if it's breaking change. Need to check it. But anyway it's good change.

Leonya:
Released in pre.145

fluidsonic:
What's the reason for the `Any` requirement on `RRef` and the nullable `current`?

My implementation:

```kt
external interface RRef<out Value> {
    val current: Value
}

external interface RMutableRef<Value> : RRef<Value> {
    override var current: Value
}
```

```kt
fun <Value : Any> useRef(): RMutableRef<Value?>
fun <Value> useRef(initialValue: Value): RMutableRef<Value>
```

-> 

```kt
val ref: RMutableRef<String> = useRef("default")
val value: String = ref.current // totally fine and non-null
```

Typescript definitions:

```ts
interface MutableRefObject<T> {
    current: T;
}

function useRef<T>(initialValue: T): MutableRefObject<T>;
```

Skolotsky:
@fluidsonic , you are right

turansky:
@fluidsonic I don't see any correlation between non-nullable `initialValue` and non-nullable `current`. 
Do you mean that "initial" == "default"?

fluidsonic:
I'm not sure what you mean.

The following should work with current implementation but doesn't because `current` is always nullable:
```kt
val ref: RMutableRef<String> = useRef("default")
val value: String = ref.current
```

turansky:
@fluidsonic [Following example](https://codesandbox.io/s/zealous-swartz-5okbf) describe that `current` can be `null` even if `initialValue` is non-null

fluidsonic:
@turansky no initial value is passed in that example. `defaultRef` is always `undefined`.

turansky:
[Fixed](https://codesandbox.io/s/friendly-sammet-nl9yx). After 2 button clicks you will see `null` current value in console

fluidsonic:
@turansky It's still broken. I've added the missing `return` myself.

I see what you mean. However that doesn't mean that `T` should be `Any`. `current` can still be non-null.
It's just in that specific *use* of the Ref its argument must be defined as nullable.

In your example you'd need an explicit nullable type argument:
```kt
val defaultRef /* : HTMLElement */ = useMemo { … } // okay
val ref1 /* : RMutableRef<HTMLElement> */ = useRef(defaultRef) // wrong
val ref1: RMutableRef<HTMLElement?> = useRef(defaultRef) // correct
val ref1 /* : RMutableRef<HTMLElement?> */ = useRef<HTMLElement?>(defaultRef) // correct
```

And `ButtonProps` would have a definition like this:
```kt
external interface ButtonProps { // whatever it's called
    val ref: RMutableRef<in HTMLButtonElement?>?  // explicitly defines it as nullable
}
```

In most other cases `(Mutable)Ref<T>` is merely another name for `Box<T>`. Boxes don't require non-null type arguments. Only some use cases do.

fluidsonic:
Just found the following in my project:

```kt
val ignoresClicks = useRef(false)
val mouseIsDownOnDropdownItem = useRef(false)
val scrollItemsIntoViewIsPending = useRef(false)
```

Ref is used to add non-rerendering mutable state locally. In this case a non-nullable `Boolean`.

turansky:
@Skolotsky, @fluidsonic I update "ref" contracts in #413

1. Contracts are less safe, but more useful
2. Unexpected `null` current value can be detected in some corner cases ([like this](https://codesandbox.io/s/blissful-beaver-wy7fo))

fluidsonic:
It's not a contract. It's simply the purpose of `Ref`. It wraps any value of any type, including `null`, in a persistent JS object.
Your example is completely null-safe in Kotlin if the type of the Button's `ref` is specified correctly. The user cannot get that wrong without an unchecked cast 🤔

Be aware that when un-marking `current` as nullable some other API of `kotlin-react` that makes use of `RMutableRef` or `RReadableRef` may now need to specify nullability on the use-site.

turansky:
> if the type of the Button's `ref` is specified correctly

This "if" means "less safe"

fluidsonic:
@turansky, not, it the responsibility of `kotlin-react` maintainers to properly define that property because the library is interfacing directly with JS code. It's no different than using any other external code and nullability is nothing special here. You could also write `Int` instead of `RMutableRef`.

turansky:
@fluidsonic I checked your cases
Looks like best variant for now - save current "safe" behaviour (with nullable `current`)
Additionally we can add following extension
```Kotlin
val <T: Any> RReadableRef<T>.currentValue: T
    get() = requireNotNull(current)
```

#413 cancelled as unsafe

cc @Skolotsky

turansky:
@fluidsonic For ref values I use following delegate:
```Kotlin
fun <T> useRefValue(
    initialValue: T,
): ReadWriteProperty<Nothing?, T> =
    RefDelegate(useRef(initialValue))

private class RefDelegate<T>(
    private val ref: RMutableRef<T>,
) : ReadWriteProperty<Nothing?, T> {
    override fun getValue(
        thisRef: Nothing?,
        property: KProperty<*>,
    ): T {
        return ref.current
    }

    override fun setValue(
        thisRef: Nothing?,
        property: KProperty<*>,
        value: T,
    ) {
        ref.current = value
    }
}
``` 

After #424 it will contains `unsafeCast` in `getValue`

fluidsonic:
@turansky why not just add delegation support to `RReadableRef` and `RMutableRef` by adding (inlined) delegate operator functions as extensions? No need for all that overhead.

In JS size and speed matter when targeting the browser.

turansky:
> In JS size and speed matter when targeting the browser.

@fluidsonic You are right. In my cases I have no many refs. 
But after migration on IR I will use following: 
```Kotlin
external interface RefDelegate<T>: ReadWriteProperty<Nothing?, T> {
    final override inline fun getValue(
        thisRef: Nothing?,
        property: KProperty<*>,
    ): T {
        return asDynamic().current
    }

    final override inline fun setValue(
        thisRef: Nothing?,
        property: KProperty<*>,
        value: T,
    ) {
        asDynamic().current = value
    }
}
```

fluidsonic:
@turansky I don't think that an `external interface` can (nor should) extend the regular `interface ReadWriteProperty`. Also, specifying that interface is not needed.

The whole purpose of `RReadableRef`/`RMutableRef` is already to be a delegate. There's no need to wrap or cast them.
You just need to provide Kotlin's delegate operators so that they can be used with `by` delegation.

So it can be simplified to:

```kt
public inline operator fun <Value> RReadableRef<Value>.getValue(thisRef: Any?, property: KProperty<*>): Value =
	current

public inline operator fun <Value> RMutableRef<Value>.setValue(thisRef: Any?, property: KProperty<*>, value: Value) {
	current = value
}
```

I'd also use `Any?` over `Nothing?` for `thisRef` as there's no reason to limit the scope.

PR #413, opened 2021-02-15
Fix "useRef" signature

turansky:
[Test example](https://codesandbox.io/s/zealous-swartz-5okbf)

Second signature - [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a7de57784aa1990d0849db96157acd960b87fa5c/types/react/index.d.ts#L1060)


Skolotsky:
What about https://github.com/JetBrains/kotlin-wrappers/pull/412#issuecomment-779261474 ?

turansky:
Cancelled as unsafe

PR #418, opened 2021-02-18
InnerHTML - plain object. Remove redundant anonymous class

turansky:
cc @SebastianAigner


SebastianAigner:
Good choice. We had actually discussed this after my PR as well, (for now) this makes for a prettier solution. Thanks for the contribution!

Leonya (approved):

Leonya:
Released in pre.147

PR #424, opened 2021-02-26
Fix ref object contract

turansky:
### Problems
1. Non-strict contract for `current`. `current` must be nullable
2. Non-strict type parameter for reference types (result of problem 1)

### Example 
[Test project](https://codesandbox.io/s/blissful-beaver-wy7fo)
After second button click one ref is `null` (see console)

### Migration
* `RReadableRef<MyElement?>` -> `RReadableRef<MyElement>`
* `RMutableRef<MyElement?>` -> `RMutableRef<MyElement>`
* `useRef<MyElement?>` -> `useRef<MyElement>`
* `createRef<MyElement?>` -> `createRef<MyElement>`

cc @vlsi


Leonya:
Released in pre.155

PR #428, opened 2021-03-17
Add Official Gradle Wrapper Validation Action

hfhbd:
See: https://github.com/gradle/wrapper-validation-action


Leonya:
Released in pre.150

PR #438, opened 2021-04-28
react-table. IR compatibility fixes

turansky:
Related issue - [KT-46364](https://youtrack.jetbrains.com/issue/KT-46364)


Leonya:
Released in pre.154

Issue #431, opened 2021-04-03
Publish to Maven Central?


joffrey-bion:
It is inconvenient to have to declare 2 different custom repos to consume even just one of the dependencies (kotlin-js-wrappers + kotlinx-html maven repos on Space).

Is there any plan to publish all of this in the same place? (preferably maven central)

NiematojakTomasz:
https://kotlin.jetbrains.space/ is not even public...

joffrey-bion:
@NiematojakTomasz The Space server itself is not public, but the maven package repository inside is in fact accessible publicly.

You have to go directly to the maven repository:
https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-js-wrappers/org/jetbrains/

For `kotlinx-html` it's a different maven repository:
https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven

turansky:
`kotlinx-html` now available in Maven Central

joffrey-bion:
@turansky Awesome! Do you know if there are plans for kotlin js wrappers to follow suit?

turansky:
[Related Slack discussion](https://kotlinlang.slack.com/archives/C5ZTZ6ER0/p1615980908000700?thread_ts=1615911674.010100&cid=C5ZTZ6ER0)

Leonya:
Waiting for approval: https://issues.sonatype.org/browse/OSSRH-69043

Leonya:
Finally: https://repo1.maven.org/maven2/org/jetbrains/kotlin-wrappers/

search.maven.org doesn't list the new packages yet, I am told it takes a while for it to update.

Note that group ID was changed from `org.jetbrains` to `org.jetbrains.kotlin-wrappers` and the version number was bumped to `pre.200+` to reflect this change. Please refer to README.md for more details.

PR #442, opened 2021-05-12
Fix `groupId`

turansky:
Sync group naming with other JetBrains libraries (Kotlin, Coroutines, Serialization)

cc @ilgonmic, @SebastianAigner, @kalibrov520

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-group=org.jetbrains
+group=org.jetbrains.kotlin-wrappers
 version=1.0-SNAPSHOT
 version.build=pre.155
 


ilgonmic:
Why do we need to broke users, which rely on current coordinates?
Users which want to update version need to know, that group was updated too, and it is very error-prone

turansky:
> Why do we need to broke users, which rely on current coordinates?

Main goal - fix `groupId` according to [maven guide](https://maven.apache.org/guides/mini/guide-naming-conventions.html) before Maven Central publish (#431)

Unique `groupId` is important for Gradle/Maven configuration, search, analysis...  
Kotlin, Coroutines and Serialization use unique `groupId`. 

> Users which want to update version need to know, that group was updated too

Yes, announcement will be required (cc @SebastianAigner)

vlsi:
Technically speaking, Maven has `relocation` feature, so if you publish the artifact under old `groupId` for a while, it would automatically redirect and warn users: https://maven.apache.org/guides/mini/guide-relocation.html

```xml
The minimal POM file might look like this for version 1.0 of foo:

<project>
  <modelVersion>4.0.0</modelVersion>
  <groupId>bar</groupId>
  <artifactId>foo</artifactId>
  <version>1.0</version>
  <distributionManagement>
    <relocation>
      <groupId>org.bar</groupId>
    </relocation>
  </distributionManagement>
</project>
```

PR #455, opened 2021-05-29
`renderToString` and `renderToStaticMarkup` functions are supported

sgrishchenko:
`renderToString` and `renderToStaticMarkup` functions are supported
https://reactjs.org/docs/react-dom-server.html


Leonya:
Released in pre.205

PR #464, opened 2021-06-01
Fix link to 'Hooks and Function Components'

aerialist7:


aerialist7:
Old links works incorrect
image

PR #467, opened 2021-06-03
Remove possibilty to override props

ilgonmic:
`props` instantiated inside React, the problem that `open val` allow to override with `override var`, and setter will be overridden.

But when during native react constructor, react set props, it calls overridden setter, which is not good, I guess


Leonya:
Released in pre.207

Issue #463, opened 2021-05-31
Consistency of hook dependencies 


sgrishchenko:
Now there are three ways to use dependencies of hooks:
1) Array (dependencies are last): `useCallback({ value }, arrayOf(value))`
1) Vararg (dependencies are first): `useCallback(value) { value }`
1) List (dependencies are first): `useEffect(listOf(value)) { println(value) }`

I suggest to use only one form - Vararg (dependencies are first). I think it is most compact form and Kotlin friendly way (because callbacks are last parameter in this case).

I can create PR, where I will add overloads for this kind of API and mark all other form as deprecated for backward compatibility.

Leonya:
@Skolotsky @turansky WDYT?

Skolotsky:
i'm for vararg version

turansky:
`useEffect` + `useLayoutEffect` - last non-updated hooks

I use/like vararg version only :)

```
inline fun useEffectOnce(
    noinline effect: () -> RCleanup,
) {
    rawUseEffect(effect, emptyArray())
}

inline fun useEffect(
    vararg dependencies: dynamic,
    noinline effect: () -> RCleanup,
) {
    rawUseEffect(effect, dependencies)
}

// super draft name
inline fun useSimpleEffect(
    vararg dependencies: dynamic,
    noinline effect: () -> Unit,
) {
    rawUseEffect(effect, dependencies)
}
```

Sync with `useLayoutEffect` - open question :(

cc @vlsi

turansky:
Also [`EffectBuilder`](https://github.com/fluidsonic/fluid-react/blob/968a473bb5b0f86ae98f757aac2835841f00a2c9/modules/playground/sources-js/EmojiContainer.kt#L12) can be used for single hook declaration:

```Kotlin
var count by useState(3)

useEffect(count) {
    val timerId = window.setTimeout({ count += 1 }, 2000)

    cleanup { window.clearTimeout(timerId) }
}
```

With effect builder we will have:
1. `useEffect()` - every render
2. `useEffectOnce()` - one time (from [`react-use`](https://github.com/streamich/react-use/blob/master/docs/useEffectOnce.md))
3. `useEffect(a, b, c)` - on change

Without effect builder we will have `x2` methods :(

Leonya:
@sgrishchenko please create a PR

sgrishchenko:
[...]

PR #468, opened 2021-06-04
Consistency of hook dependencies

sgrishchenko:
All previous effect and memoization hooks are deprecated with suggestions and replacements, new ones are provided.
See #463


turansky:
1. Where is varargs? :)
2. `EffectBuilder` must be synthetic class (`Array` in practice) with `inline` methods only

Skolotsky:
Yes, my vote for varargs

PR #470, opened 2021-06-04
Add `REffectBuilder`

### LINKED ISSUE #463: Consistency of hook dependencies

turansky:
Fix #463. Part 1

### DIFF
--- a/kotlin-react/src/main/kotlin/react/REffectBuilder.kt
+++ b/kotlin-react/src/main/kotlin/react/REffectBuilder.kt
@@ -1,0 +1,34 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
+package react
+
+// TODO: make external in IR
+class REffectBuilder
+private constructor() {
+    inline fun cleanup(
+        noinline block: RCleanup,
+    ) {
+        asDynamic().push(block)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


turansky:
@sgrishchenko Draft for part 2:
```Kotlin
fun useEffectOnce(
    effect: REffectBuilder.() -> Unit,
) {
    val callback = useEffectCallback(effect)
    rawUseEffect(callback, emptyArray())
}

fun useEffect(
    vararg dependencies: dynamic,
    effect: REffectBuilder.() -> Unit,
) {
    val callback = useEffectCallback(effect)
    rawUseEffect(callback, dependencies)
}

```

Leonya:
Released in pre.208

PR #472, opened 2021-06-05
Add `useRefValue` hook

turansky:
```Kotlin
var count by useRefValue(0)

// debug mode
useEffect {
    count++
    println(count)
}
```


Leonya:
Released in pre.208

PR #476, opened 2021-06-07
Consistency of hook dependencies v2

### LINKED ISSUE #463: Consistency of hook dependencies

sgrishchenko:
Fix #463
New implementation of hooks with vararg dependencies is added, old implementation is deprecated.
Cc @turansky

### DIFF
--- a/kotlin-react/src/main/kotlin/react/useEffect.kt
+++ b/kotlin-react/src/main/kotlin/react/useEffect.kt
@@ -12,8 +12,8 @@ typealias RCleanup = () -> Unit
 fun useEffect(
     effect: EffectBuilder.() -> Unit,
 ) {
-    val effectCallback = useEffectCallback(effect)
-    rawUseEffect(effectCallback)
+    val callback = useEffectCallback(effect)
+    rawUseEffect(callback)
 }
 
 /**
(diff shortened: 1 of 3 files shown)


kotlin-react/src/main/kotlin/react/useEffect.kt
```diff
@@ -9,6 +9,43 @@ typealias RCleanup = () -> Unit
  * Only works inside [functionalComponent]
  * @see <a href="https://reactjs.org/docs/hooks-state.html#hooks-and-function-components">Hooks and Function Components</a>
  */
+fun useEffect(
+    effect: EffectBuilder.() -> Unit,
+) {
+    val effectCallback = useEffectCallback(effect)
```
turansky: `callback`?

turansky (approved):

kotlin-react/src/main/kotlin/react/useEffect.kt
sgrishchenko: fixed

turansky (approved):

turansky:
And special thanks you to @vlsi and @fluidsonic for `EffectBuilder` idea

Leonya:
Released in pre.209

PR #478, opened 2021-06-09
Redundant `useReducer` wrappers are removed

sgrishchenko:
1) `ReducerInstance` is provided to reduce generated js code
1) Redundant `useReducer` wrappers are removed

сс @turansky


kotlin-react/src/main/kotlin/react/Imports.kt
```diff
@@ -110,20 +110,28 @@ external val Profiler: RClass<ProfilerProps>
...
-    valueInitializer: () -> T,
+    initializer: () -> T,
 ): StateInstance<T>
 
 // Reducer Hook (16.8+)
-@JsName("useReducer")
-external fun <S, A> rawUseReducer(
+external fun <S, A, I> useReducer(
+    reducer: RReducer<S, A>,
+    initializerArg: I,
```
turansky, line 123: `initialArg` in original declarations?

turansky (approved):

kotlin-react/src/main/kotlin/react/Imports.kt
sgrishchenko: I have used names from [Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L988), I like their naming more because of clarity

Leonya:
Released in pre.209

PR #480, opened 2021-06-11
Fix `PropGetter` contract

turansky:
Use [most typical contract](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/51e77dacacf9d22ede03f0c7beeb244dc5346f1a/types/react-table/index.d.ts#L182)


Leonya:
Released in pre.210

PR #483, opened 2021-06-12
Fix typo in GridAutoRows and GridAutoColumns

nordiauwu:
https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-rows#syntax
https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-columns#syntax


Leonya:
Released in pre.210

PR #494, opened 2021-06-21
Update `redux` dependencies

turansky:


Leonya:
Released in pre.213-kotlin-1.5.20

PR #497, opened 2021-06-24
jsStyle not written into string attributes

ilgonmic:
`kotlinx.html` has `attributes` map `String` -> `String`
So, it is't allowed to pit dynamic here.
In legacy backend, there is no check in this place (but it is incorrect action in Kotlin type system)
But in IR there is type check, and it breaks on runtime


Leonya:
Released in pre.213-kotlin-1.5.20

PR #501, opened 2021-07-06
React Query Declarations

aerialist7:
cc @turansky


gradle.properties
```diff
@@ -26,6 +26,9 @@ react.version=17.0.2
 # kotlin-react-dom
 react-dom.version=17.0.2
 
+# kotlin-react-query
+react-query.version=3.18.1
```
turansky, line 30: Latest?

turansky (approved):

gradle.properties
Leonya: It is.

Leonya:
Released in pre.215

Issue #358, opened 2020-10-21
[react] Component and Builder APIs need cleanup


fluidsonic:
When I've started working with kotlin-react I've had a lot of confusion because the API is not consistent nor easy to understand.

To name a few:

- Some things work with class-based components but not with functional components and vice versa.
- The provided types are inconsistently named and their hierarchy confusing.
- The relations between and intent of the various types are confusing.
- Some functionality was quite difficult to implement, e.g. using `styled` as HOC for functional components.

### Status quo

I'll go through a collection of inconsistencies that I've noticed.

There are two functions to create functional components that are almost identical:

https://github.com/JetBrains/kotlin-wrappers/blob/10805ef35995ec6a40a7e006844b0d44b4ce3723/kotlin-react/src/main/kotlin/react/RBuilder.kt#L176-L189

https://github.com/JetBrains/kotlin-wrappers/blob/10805ef35995ec6a40a7e006844b0d44b4ce3723/kotlin-react/src/main/kotlin/react/ReactComponent.kt#L84-L91

A crucial difference is that one returns `RClass<P>` while the other returns `FunctionalComponent<P>`.

- From React's point of view, both are component definitions.
- From Kotlin's point of view, `RClass<P>` and `FunctionalComponent<P>` have no common interface and thus cannot be used interchangeably where a component definition is expected.
- In `rFunction` above the `FunctionalComponent` is even cast to `RClass<P>`.
- `RClass` has the omnipresent `R` prefix while `FunctionalComponent` does not.

That makes it difficult to use API that is made for either one or the other but could easily be used for both, for example:

- `HOC` operates on `RClass`.
- `lazy/rLazy()` operate on `RClass`.
- `styled()` operates on `RClass`.
- `memo()` operates on `FunctionalComponent`.

Then there's also `RBuilder.child()` which is inconsistent across overloads:

- One operates on `Child`, i.e. not a component but an actual element. The very generic name `Child` is confusing because components are also added as children, but aren't a `Child`.
- One operates on `Any` and is type-unsafe. `handler` is a required parameter. `props` is a **required** parameter.
- One operates on `KClass<out Component>`. `handler` is a required parameter. No `props` parameter.
- One operates on `reified Component`. `handler` is a required parameter. No `props` parameter.
- One operates on `FunctionalComponent`. `handler` is an **optional** parameter. `props` is an **optional** parameter. It's an **extension** function.
- **None** operates on `RClass` explicitly.

[...]

PR #509, opened 2021-07-09
All RBuilder child declarations are extracted to separate file

sgrishchenko:
These changes are related to #358


turansky (commented):
In any case we need cleanup before
I suggest close this PR (without merge) and discuss changes one more time

PR #514, opened 2021-07-10
#513 Remove deprecated effect hooks

### LINKED ISSUE #513: Can't pass a single list as a dependency of a useEffect without using the deprecated function
[Slack thread](https://app.slack.com/client/T09229ZC6/C5ZTZ6ER0/thread/C5ZTZ6ER0-1625668819.023000)

To sum up, until now I was using:
```kotlin
    useEffect(listOf(props.imageUrls)) {
          //My effect
    }
```

This way I was not checking individual elements of `props.imageUrls` in my `useEffect` but I was only checking if the list in itself has changed (which was useful for my specific case).

Since the deprecation of the previous usage, I would like to use the non-deprecated `useEffect`:
```kotlin
    useEffect(props.imageUrls) {
          //My effect
    }
```
However by doing the above, the deprecated function is called (which is not equivalent and not what I want).

As suggested by @turansky, I could use `rawUseEffect` to create a custom wrapper. However, I was wondering if the deprecated functions could not simply be removed (since breaking changes have been introduced any way with the `EffectBuilder`).

turansky:
Fix #513

cc @sgrishchenko


Leonya:
Released in pre.216

PR #516, opened 2021-07-10
Kotlin CSS: Declare `isolation` property

aerialist7:
https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

cc @turansky

### DIFF
--- a/kotlin-css/src/commonMain/kotlin/kotlinx/css/StyleEnums.kt
+++ b/kotlin-css/src/commonMain/kotlin/kotlinx/css/StyleEnums.kt
@@ -2,7 +2,10 @@
 
 package kotlinx.css
 
-import kotlin.math.*
+import kotlin.math.PI
+import kotlin.math.abs
+import kotlin.math.max
+import kotlin.math.roundToInt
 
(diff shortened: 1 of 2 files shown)


kotlin-css/src/commonMain/kotlin/kotlinx/css/StyleEnums.kt
```diff
@@ -38,6 +41,16 @@ enum class Appearance {
...
 }
 
+enum class Isolation {
+    initial, inherit, revert, unset,
+
+    auto, isolate,
+
+    ;
+
+    override fun toString() = name.hyphenize()
```
Leonya, line 51: Not needed as none of the values require hyphenization.

Leonya (changes requested):

kotlin-css/src/commonMain/kotlin/kotlinx/css/StyleEnums.kt
aerialist7: Fixed

Leonya:
Released in pre.216

PR #547, opened 2021-07-17
`RBuilder.child` returns `Unit`

sgrishchenko:
# !!!WARNING: BREAKING CHANGE!!!
I recommend to increase version to `pre.300` to notify users that these changes are important. 

### Explanation
**React Original implementation:** render is element factory with signature: `(RProps) -> ReactElement`
**Kotlin Wrappers implementation:** render is builder invocation with signature: `RBuilder.(RProps) -> Unit`

`RBuilder.child` had strange signature `RBuilder.(RProps) -> ReactElement`.
 
As result controversial API led to strange side effects (e.g. `RBuilder.route` children were added to parent `RBuilder` and returned `ReactElement` was passed to child `RouteComponent` at the same time).

### Additional change
`RBuilder.route` receives `RRender` instead of `ReactElement` factory

### DIFF
--- a/kotlin-react/src/main/kotlin/react/RBuilder.kt
+++ b/kotlin-react/src/main/kotlin/react/RBuilder.kt
@@ -14,9 +14,8 @@ annotation class ReactDsl
 interface RBuilder {
     val childList: MutableList<Any>
 
-    fun <T : Child> child(element: T): T {
+    fun <T : Child> child(element: T) {
         childList.add(element)
-        return element
     }
 
     operator fun Child.unaryPlus() {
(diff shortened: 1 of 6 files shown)


kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -31,35 +30,35 @@ interface RBuilder {
         type: ComponentType<P>,
         props: P,
         children: List<Any>,
-    ): ReactElement =
+    ) =
```
turansky, line 33: Function body?

kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -31,35 +30,35 @@ interface RBuilder {
...
             childList
         }
-        return child(type, props, children)
+        child(type, props, children)
     }
 
     operator fun <P : RProps> ComponentType<P>.invoke(
         handler: RHandler<P>,
-    ): ReactElement =
+    ) =
```
turansky, line 50: Function body?

kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -73,20 +72,20 @@ interface RBuilder {
     fun <P : RProps> RClass<P>.node(
         props: P,
         children: List<Any> = emptyList(),
-    ): ReactElement =
+    ) =
```
turansky, line 75: Function body?

turansky (changes requested):

turansky:
@sgrishchenko Router signature was invalid, it also can be separate change

PR #548, opened 2021-07-17
Ring UI: `RClass` -> `ComponentClass`

aerialist7:
cc @altavir @turansky


altavir (commented):
I do not know what is the difference between RClass and ComponentClass, but no complaints here.

aerialist7:
> I do not know what is the difference between RClass and ComponentClass, but no complaints here.

`RClass` will be deprecated and deleted soon

turansky (approved):

Leonya:
Released in pre.218

PR #551, opened 2021-07-17
New FC signatures don't receive nullable `displayName`

sgrishchenko:
cc @turansky @bnorm


kotlin-react/src/main/kotlin/react/FunctionalComponent.legacy.kt
```diff
@@ -15,5 +15,9 @@ typealias FunctionalComponent<P> = FunctionComponent<P>
 inline fun <P : RProps> functionalComponent(
```
turansky, line 15: Redundant?

turansky (changes requested):

turansky (approved):

Leonya:
Released in pre.218

PR #561, opened 2021-07-18
`RRender` -> `Render`

aerialist7:
cc @turansky @sgrishchenko

### DIFF
--- a/kotlin-react/src/main/kotlin/react/RBuilder.kt
+++ b/kotlin-react/src/main/kotlin/react/RBuilder.kt
@@ -7,12 +7,6 @@ import kotlin.reflect.KClass
 
 typealias Render = RBuilder.() -> Unit
 
-@Deprecated(
-    message = "Will be removed soon",
-    replaceWith = ReplaceWith("Render", "react.Render"),
-)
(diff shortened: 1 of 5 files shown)


kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -5,7 +5,13 @@ import kotlinext.js.js
...
 import kotlin.reflect.KClass
 
-typealias RRender = RBuilder.() -> Unit
+typealias Render = RBuilder.() -> Unit
+
+@Deprecated(
+    message = "Will be removed soon",
+    replaceWith = ReplaceWith("Render", "react.Render"),
+)
+typealias RRender = Render
```
turansky, line 14: Redundant?

turansky (changes requested):

kotlin-react/src/main/kotlin/react/RBuilder.kt
aerialist7: Fixed

Leonya:
Released in pre.219

PR #566, opened 2021-07-20
Add interface for CSSBuilder injecting directly in styled components.

tretikoff:
Add interface for CSSBuilder injecting directly in styled components. It is needed because CSSBuilder contains already parsed rules, with which we can work directly in CSSOM later


Leonya:
Released in pre.220

PR #570, opened 2021-07-24
Use `ComponentClass` instead of `RClass`

aerialist7:
cc @turansky @sgrishchenko

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/helpers.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/helpers.kt
@@ -3,8 +3,8 @@ package react.router.dom
 import react.*
 import kotlin.reflect.KClass
 
-fun <P : RProps> withRouter(klazz: KClass<out Component<P, *>>): RClass<P> =
+fun <P : RProps> withRouter(klazz: KClass<out Component<P, *>>): ComponentClass<P> =
     rawWithRouter(klazz.rClass)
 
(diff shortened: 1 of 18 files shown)


turansky (approved):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/helpers.kt
```diff
@@ -3,8 +3,8 @@ package react.router.dom
 import react.*
 import kotlin.reflect.KClass
 
-fun <P : RProps> withRouter(klazz: KClass<out Component<P, *>>): RClass<P> =
+fun <P : RProps> withRouter(klazz: KClass<out Component<P, *>>): ComponentClass<P> =
```
sgrishchenko, line 6: I think we need some analysis for these changes. For example, it seems we should use `ComponentType` here because we should not know how `withRouter` HOC implemented. It can return `FunctionComponent` under the hood, but we are saying that it is Class. I recommend use `ComponentType` everywhere it is possible in libraries for returned types and use `ComponentClass` only if it is really needed.

sgrishchenko (changes requested):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/helpers.kt
turansky: Looks like next PR (check required before)

Leonya:
Released in pre.222

PR #575, opened 2021-07-27
Remove unsafe `childFunction`

sgrishchenko:
Part of #358
cc @turansky @aerialist7 

`childFunction` and `RProps.children(value: T)` signatures are removed because of unsafety.
These methods can lead to inconsistency and mistakes in codebase of `kotlin-wrappers` users.

Recommended solution for such cases:
```
interface RenderChildrenProps : RProps {
    var children: RBuilder.(value: Int) -> Unit
}

val RenderChildren = fc<RenderChildrenProps> { props ->
    div {
        props.children(this, 42)
    }
}

val Container = fc<RProps> {
    RenderChildren {
        attrs {
            children = { value ->
                span {
                    attrs {
                        jsStyle {
                            color = "red"
                        }
                    }

                    +"$value"
                }
            }
        }
    }
}
```

### DIFF
--- a/kotlin-react/src/main/kotlin/react/RBuilder.kt
+++ b/kotlin-react/src/main/kotlin/react/RBuilder.kt
@@ -87,18 +87,6 @@ interface RBuilder {
         klazz.react.invoke(handler)
     }
 
-    fun <T, P : RProps> childFunction(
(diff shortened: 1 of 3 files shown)


turansky (approved):

Leonya:
Released in pre.223

PR #577, opened 2021-07-28
Move `Key` from `react-table` to `react`

aerialist7:
cc @turansky @sgrishchenko

### DIFF
--- a/kotlin-react/src/main/kotlin/react/Key.kt
+++ b/kotlin-react/src/main/kotlin/react/Key.kt
@@ -1,3 +1,3 @@
 package react
 
-typealias ReactKey = String
+typealias Key = String
--- a/kotlin-react/src/main/kotlin/react/Key.kt
+++ b/kotlin-react/src/main/kotlin/react/Key.kt
@@ -1,0 +1,3 @@
+package react
+
+typealias ReactKey = String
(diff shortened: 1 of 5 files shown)


kotlin-react/src/main/kotlin/react/Props.kt
```diff
@@ -5,7 +5,7 @@ external interface RProps
 val RProps.children: Any
     get() = asDynamic().children
 
-var RProps.key: String
+var RProps.key: ReactKey
```
turansky, line 8: One more place?

kotlin-react/src/main/kotlin/react/Key.kt
```diff
@@ -0,0 +1,3 @@
+package react
+
+typealias ReactKey = String
```
turansky, line 3: `Key` like in types?

turansky (changes requested):

kotlin-react/src/main/kotlin/react/Key.kt
aerialist7: Done

kotlin-react/src/main/kotlin/react/Props.kt
aerialist7: OK

turansky (approved):

Leonya:
Released in pre.223

PR #581, opened 2021-07-30
Isolation for `css` block (no DOM in `css`)

turansky:
cc @sgrishchenko @aerialist7 

Screenshot 2021-07-30 at 19 28 19


Leonya:
Released in pre.223

PR #582, opened 2021-07-30
DSL marker for `RProps`

turansky:
cc @sgrishchenko @aerialist7


Leonya:
#581 broke all kinds of things in Space as it started interfering with our own annotations, so we need to analyse the impact of this PR beforehand. @Skolotsky

turansky:
Let's start from separate file for `@ReactDsl`

Leonya:
Released in pre.224

PR #585, opened 2021-07-31
`CSSBuilder` -> `CssBuilder`

aerialist7:
cc @turansky @sgrishchenko


kotlin-css/src/commonMain/kotlin/kotlinx/css/CssBuilder.kt
```diff
@@ -1,10 +1,6 @@
 package kotlinx.css
 
-typealias RuleSet = CSSBuilder.() -> Unit
-
-fun ruleSet(set: RuleSet) = set
-
-class CSSBuilder(
+class CssBuilder(
```
turansky, line 3: `CSSBuilder` alias required (with `@Deprecate`)

turansky (changes requested):

kotlin-css/src/commonMain/kotlin/kotlinx/css/CssBuilder.kt
aerialist7: Fixed

turansky (approved):

Leonya:
I'm sorry, it's a useless change and I'm not accepting it.

aerialist7:
> I'm sorry, it's a useless change and I'm not accepting it.

Hm... Why useless? There is a coding convention rule for such cases.

https://developer.android.com/kotlin/style-guide#camel_case

Leonya:
Yes, sure there is. I didn't mind renaming `RState` and such because we aren't using them too much directly. `CSSBuilder` is a different story, it has hundreds of usages in our codebase and I simply don't want to rename them because of coding conventions. Some things are named imperfectly, let's just accept that. I have had this discussion with @turansky already and I can repeat my opinion: I appreciate your contributions, but I would really love to see a different ratio of useful, meaningful changes to "making the codebase pretty" changes. Right now we have maybe one or two useful changes per dozen of those that belong to the second category. Again, I do appreciate your contributions but you have to understand me as well: I am spending considerable time verifying the changes you've made and adapting my codebase to not have any deprecated symbols. Do I feel that I'm spending my time well? Well, I'm very sorry, but I just don't. If you want to create "perfect" Kotlin wrappers for JS libraries then maybe you should be doing this elsewhere. But if you want to be contributing to a JetBrains repository, then you need to the fact that not all of your PRs will be merged.

Leonya:
We have decided that we'll be refactoring `CSSBuilder` and `StyledElement` soon, so OK, let's rename it as well.

xian:
Uhhh... could you at least leave a deprecated `CSSBuilder` class sitting around for a while? My project depends on libraries which are now broken because of the rename. Super annoying.

xian:
@aerialist7 / @Leonya — in hopes someone will see this comment/https://github.com/JetBrains/kotlin-wrappers/pull/585#issuecomment-915581244 since I can't reopen the issue.

turansky:
@xian `CssBuilder` was changed with many other things and removed in #662 after **15** releases
Deprecated - `pre.224`
Removed - `pre.238`
 
Old `CSSBuilder` usage in library - incompatibility signal (library update required in any case)

cc @sgrishchenko

PR #588, opened 2021-07-31
DSL marker for `RProps`

turansky:
cc @sgrishchenko @aerialist7

### DIFF
--- a/kotlin-react/src/main/kotlin/react/Props.kt
+++ b/kotlin-react/src/main/kotlin/react/Props.kt
@@ -1,5 +1,6 @@
 package react
 
+@ReactDsl
 external interface RProps
 
 val RProps.children: Any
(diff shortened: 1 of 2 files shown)


aerialist7 (approved):

Leonya:
This can't be merged in a foreseeable future, see https://github.com/JetBrains/kotlin-wrappers/commit/a61e27678cd01d73d100c0671cdecc4ff2d132ac

turansky:
For now I know single root cause - invalid DSL (css in dom or dom in css)
Do you have other problems?

Leonya:
In Space we have a concept of "live" properties. These can only be used in "trackable" contexts. Once you've added these annotations, all the places where we call "Property#live" inside a `css` block (perfectly valid!) became compilation errors. This is fixable, of course, and we are fixing it already for DOM builders by introducing a parallel hierarchy of classes, along the lines of `interface XStyledDOMBuilder<out T : Tag> : StyledDOMBuilder<T>, XTrackable` but this is not an elegant fix. Right not we don't want to add these hacks for CSS builders. Once Kotlin has multiple receivers this will become easy.

turansky:
[Related Slack discussion](https://kotlinlang.slack.com/archives/C5ZTZ6ER0/p1628078963002700)

PR #590, opened 2021-08-02
Add missed `ElementType`

turansky:
[DT](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8d4c8f0a784733c050debdaa072920956a42acc4/types/react/index.d.ts#L73)

cc @sgrishchenko @aerialist7


Leonya:
Released in pre.224

PR #593, opened 2021-08-02
Ring UI Wrapper: Add Checkbox component

Neitex:


kotlin-ring-ui/src/main/kotlin/ringui/Checkbox.kt
```diff
@@ -0,0 +1,31 @@
...
+import react.dom.WithClassName
+
+external interface CheckboxProps : WithClassName {
+    var name: String
+    var label: String
+    var containerClassName: String
+    /**
+     * From React docs: "The style attribute accepts a JavaScript object with camelCased properties rather than a CSS string.".
+     */
+    var containerStyle: Object
```
turansky, line 19: `dynamic` preferred, while interface not declared

turansky (changes requested):

turansky (approved):

kotlin-ring-ui/src/main/kotlin/ringui/Checkbox.kt
turansky: `CssProperties` can be declared in comment for property for future

Leonya:
Released in pre.225

PR #598, opened 2021-08-06
Public interface tests 

tretikoff:
More tests for CSSBuilder function integration with `kotlin-styled-next`


Leonya:
Released in pre.227

PR #602, opened 2021-08-07
Prepare Component Props for more strict restrictions

sgrishchenko:
There are typing vulnerability of currents props: every component now can have children with type `Array<ReactNode>`, even it is not declared obviously. Also there is conflict of names and now there are difficulties with creation of children with custom type (e.g. render children)
This PR - first step to make `Props` more strict.
[TS typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L824)
Part of #358
cc @turansky @aerialist7

### DIFF
--- a/kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
+++ b/kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
@@ -14,7 +14,7 @@ external interface QueryErrorResetBoundaryValue {
 external val useQueryErrorResetBoundary: () -> QueryErrorResetBoundaryValue
 
 external interface QueryErrorResetBoundaryProps : react.Props {
-    var  children: dynamic
+    var children: dynamic
 }
 
 external val QueryErrorResetBoundary: react.FC<QueryErrorResetBoundaryProps>
--- a/kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
+++ b/kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
@@ -13,8 +13,8 @@ external interface QueryErrorResetBoundaryValue {
 
 external val useQueryErrorResetBoundary: () -> QueryErrorResetBoundaryValue
 
-external interface QueryErrorResetBoundaryProps : react.RProps {
-    var children: dynamic
+external interface QueryErrorResetBoundaryProps : react.Props {
+    var  children: dynamic
 }
 
 external val QueryErrorResetBoundary: react.FC<QueryErrorResetBoundaryProps>
--- a/kotlin-react/src/main/kotlin/react/RComponent.kt
+++ b/kotlin-react/src/main/kotlin/react/RComponent.kt
@@ -2,7 +2,7 @@ package react
 
 import kotlinext.js.jsObject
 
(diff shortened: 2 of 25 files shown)


kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
```diff
@@ -13,8 +13,8 @@ external interface QueryErrorResetBoundaryValue {
 
 external val useQueryErrorResetBoundary: () -> QueryErrorResetBoundaryValue
 
-external interface QueryErrorResetBoundaryProps : react.RProps {
-    var children: dynamic
+external interface QueryErrorResetBoundaryProps : react.Props {
```
turansky, line 16: `PropsWithChildren`? I will fix it in generator

kotlin-react/src/main/kotlin/react/RComponent.kt
```diff
@@ -16,10 +16,6 @@ abstract class RComponent<P : RProps, S : State> : Component<P, S> {
     // if you use this method, don't forget to pass props to the constructor first
     open fun S.init(props: P) {}
 
-    fun RBuilder.children() {
```
turansky, line 19: Redundant?

turansky (approved):

kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
```diff
@@ -13,8 +13,8 @@ external interface QueryErrorResetBoundaryValue {
...
+    var  children: dynamic
```
Leonya, line 17: Лишний пробел

turansky: I will regenerate `react-query` wrappers after this PR

kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
sgrishchenko: https://react-query.tanstack.com/reference/QueryErrorResetBoundary
There should be render children, not `Array<ReactNode>`

kotlin-react/src/main/kotlin/react/RComponent.kt
sgrishchenko: Yea, there is `context(RBuilder) Props.children()`, this declaration in more obvious, it is impossible to declare such method in `RComponent` because we don't know will inheritor implements `PropsWithChildren` or not

kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
sgrishchenko: fixed anyway

kotlin-react-query/src/main/kotlin/react/query/QueryErrorResetBoundary.kt
sgrishchenko: Fixed anyway

aerialist7 (approved):

Leonya:
Released in pre.228

PR #603, opened 2021-08-08
Safe constructor for `ReactNode` based on `String`

sgrishchenko:
Preparation for strict children type.
cc @turansky @aerialist7

### DIFF
--- a/kotlin-react/src/main/kotlin/react/ReactNode.kt
+++ b/kotlin-react/src/main/kotlin/react/ReactNode.kt
@@ -12,7 +12,7 @@ fun ReactNode.asElementOrNull(): ReactElement? = when {
     else -> null
 }
 
-fun ReactNode(string: String): ReactNode =
+inline fun ReactNode(string: String): ReactNode =
     string.unsafeCast<ReactNode>()
 
 @Deprecated(
--- a/kotlin-react/src/main/kotlin/react/ReactNode.kt
+++ b/kotlin-react/src/main/kotlin/react/ReactNode.kt
@@ -1,3 +1,5 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
 package react
 
 import kotlinext.js.asJsObject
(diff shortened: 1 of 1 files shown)


kotlin-react/src/main/kotlin/react/ReactNode.kt
```diff
@@ -12,6 +12,9 @@ fun ReactNode.asElementOrNull(): ReactElement? = when {
     else -> null
 }
 
+fun ReactNode(string: String): ReactNode =
```
turansky, line 15: `inline`?

turansky (changes requested):

kotlin-react/src/main/kotlin/react/ReactNode.kt
sgrishchenko: fixed

sgrishchenko: IDEA does not recommends use inline here
> Expected performance impact from inlining is insignificant. Inlining works best for functions with parameters of functional types
Shell I suppress it? Why do you want to avoid non-inline declaration? Do you want reduce any runtime after compilation?

turansky: > Shell I suppress it?
Yes

> Do you want reduce any runtime after compilation?
Yes, this method is just for "union emulation" (for compiler only)

turansky: [Example](https://github.com/JetBrains/kotlin-wrappers/blob/0a23d99440dca7e3e1495a5eab37c87bd0b2348a/kotlin-react/src/main/kotlin/react/TransitionInstance.kt#L1)

sgrishchenko: fixed

turansky (approved):

Leonya:
Released in pre.229

PR #604, opened 2021-08-08
Strict children

sgrishchenko:
cc @turansky @aerialist7

PR #605, opened 2021-08-08
Typesafe `createElement` and `cloneElement`

sgrishchenko:
cc @turansky @aerialist7


turansky (commented):
It will have sense only when #603 and #604 will be merged

Leonya:
Released in pre.229

PR #617, opened 2021-08-11
kotlin styled next build configuration

tretikoff:
Change build from multiplatform to js
Remove IR compiler


Leonya:
Released in pre.230

Issue #614, opened 2021-08-10
useEffectOnce / useLayoutEffectOnce / useRef broken when re-rendering at an above component


Lezzio:
Hi,

When using these hooks in a component, whenever we trigger a re-render from the above components it doesn't remember it has been fired once already and loses its "memory". Note that the re-render needs to induce DOM modifications (node add/delete as far as I know) for the bug to happen.

The expected behavior is the following on React JS :
https://codesandbox.io/s/kind-sea-45wqs?file=/src/App.tsx
But it instead acts just like this : 
https://codesandbox.io/s/kind-sea-45wqs?file=/src/App.tsx

It's what we've at least observed in our project and we reproduced the bug this way : 

**Main component with the state update and children:**
![image](https://user-images.githubusercontent.com/7536802/128907318-81cd4abc-7fa4-4cdf-b06e-fb7282b487b8.png)

**OnceComponent :** 
![image](https://user-images.githubusercontent.com/7536802/128907074-cc4597a6-3c6b-4c85-b686-d7bfce8c885b.png)

**ParentComponent :**
![image](https://user-images.githubusercontent.com/7536802/128907579-7bc92bc0-cd7f-4047-adaa-1466f897a906.png)

Console output after pressing the button multiple times : 
![image](https://user-images.githubusercontent.com/7536802/128907732-086b546d-58db-4efc-84a5-40f412a54408.png)

The minimal way to work around this issue is by using a unique key on the uppermost components next to the state being updated (direct children) : 

**Works**
![image](https://user-images.githubusercontent.com/7536802/128908413-2487905a-dd9d-477f-9b15-5884280bdf60.png)

**Works**
![image](https://user-images.githubusercontent.com/7536802/128908340-35e171bb-f634-46e1-8fa7-a203f532c5d0.png)

NB : Slack thread on the issue https://kotlinlang.slack.com/archives/C5ZTZ6ER0/p1628553043020800

turansky:
@Lezzio @PeredurOmega Do you use function components?

PeredurOmega:
Almost exclusively

sgrishchenko:
This issue is reopened because we still have problems with keys in wrappers and we have plans to introduce some solution.

turansky:
Workarounds unlocked by #1264

turansky:
Next step - compiler plugin! :)

turansky:
[...]

PR #621, opened 2021-08-11
Implementing conditional rendering

PeredurOmega:
As explained in the React documentation (https://reactjs.org/docs/conditional-rendering.html), Kotlin JS should provide conditional rendering (like in JSX). This should fix https://github.com/JetBrains/kotlin-wrappers/issues/614 (following [Slack thread](https://kotlinlang.slack.com/archives/C5ZTZ6ER0/p1628553043020800)).

### DIFF
--- a/kotlin-react/src/main/kotlin/react/RBuilder.kt
+++ b/kotlin-react/src/main/kotlin/react/RBuilder.kt
@@ -156,6 +156,20 @@ interface RBuilder {
     }
 
     fun ReactElement.withKey(newKey: Number) = withKey(newKey.toString())
+
+    /**
+     * Allow conditional rendering (the same way JSX does)
+     * If the [condition] is true a [Fragment] with the children present in the [handler] is added to the current
+     * [childList] else null is added to the current [childList].
+     * https://reactjs.org/docs/conditional-rendering.html
+     */
+    private inline fun conditional(condition: Boolean, handler: Render) {
+        val childFragment = if (condition) {
+            val childList = RBuilder().apply(handler).childList
+            createElement(Fragment, jsObject(), *childList.toTypedArray())
+        } else createElement(Fragment, jsObject(), null)
+        childList.add(childFragment)
+    }
 }
 
 @JsName("createBuilder")


kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -156,6 +156,20 @@ interface RBuilder {
...
 
     fun ReactElement.withKey(newKey: Number) = withKey(newKey.toString())
+
+    /**
+     * Allow conditional rendering (the same way JSX does)
+     * If the [condition] is true a [Fragment] with the children present in the [handler] is added to the current
+     * [childList] else null is added to the current [childList].
+     * https://reactjs.org/docs/conditional-rendering.html
+     */
+    private inline fun conditional(condition: Boolean, handler: Render) {
```
Leonya, line 166: Why `private`? Not sure about `inline` either.

turansky:
@PeredurOmega 
Right now we refactor `RBuilder` and related classes. 
New `RBuilder` methods not planed

1. `conditional` does't look optimal
2. Now it can work outside `RBuilder` for your problems
3. After refactoring we will check #614 one more time

cc @sgrishchenko 

P.S. `conditional ` implementation can be simplified

kotlin-react/src/main/kotlin/react/RBuilder.kt
PeredurOmega: My bad, it should not be private. And idk about inline because its behavior seems 'unexpected' to me.

PeredurOmega:
Ok thanks @turansky for your comment. Do you have a timeline in mind for the refactoring of RBuilder ?

turansky:
> Do you have a timeline in mind for the refactoring of RBuilder ?

No strict deadlines
The end of September - good time for ping in #614

sgrishchenko:
`conditional` looks like a narrow solution, thare are also `when` operator, also people often use `?. let { }` to implement conditional rendering for nullable values, so if we want to add some solution in library I think it should be more general. But thank you for affords.

PeredurOmega:
@sgrishchenko I totally agree with you, a broader solution would be great (especially because it would be an improvement over JSX). Are you considering handling this during compilation or have you other ideas in mind ?
In the meantime I will continue to use this for my personal use to avoid unneeded rendering.

sgrishchenko:
Yes, I and @turansky discussed compiler plugin or something like that, but we don't have concrete plans right now

turansky:
#614 - more right place for related discussions :)

turansky:
[Compiler plugin](https://github.com/turansky/seskar#conditional-rendering) on board!

PR #623, opened 2021-08-12
React events [generated]

turansky:

cc @sgrishchenko


Leonya:
Released in pre.230

PR #624, opened 2021-08-12
Fix unexpected ambiguity

sgrishchenko:
Fix unexpected ambiguity, child signature with props and handler is returned
cc @turansky 
See also: https://youtrack.jetbrains.com/issue/KT-48228


turansky (approved):

Leonya:
Released in pre.231

PR #629, opened 2021-08-15
React ARIA unions [generated]

turansky:

cc @sgrishchenko


kotlin-react-dom/src/main/generated/react/dom/AriaChecked.kt
```diff
@@ -0,0 +1,14 @@
...
+
+package react.dom
+
+@Suppress("NAME_CONTAINS_ILLEGAL_CHARS")
+// language=JavaScript
+@JsName("""({__false__: 'false', mixed: 'mixed', __true__: 'true'})""")
+external enum class AriaChecked {
+    __false__,
+    mixed,
+    __true__,
```
sgrishchenko, line 11: If you use higher case naming, you won't have to use these controversial naming for boolean values

turansky: Upper case doesn't look like better solution better solution for me

sgrishchenko: Could you show some examples and I will think what we can do? Or you just want to be aligned with TS/HTML specification?

turansky: Yes, I want to be aligned with TS
* to simplify search
* to simplify next migration (when unions will be supported in Kotlin)

Leonya:
Released in pre.232

PR #631, opened 2021-08-16
React `PointerType` [generated]

turansky:

cc @sgrishchenko

### DIFF
--- a/kotlin-react-dom/src/main/generated/react/dom/PointerType.kt
+++ b/kotlin-react-dom/src/main/generated/react/dom/PointerType.kt
@@ -1,0 +1,14 @@
+// Automatically generated - do not modify!
+
+package react.dom
+
+@Suppress("NAME_CONTAINS_ILLEGAL_CHARS")
+// language=JavaScript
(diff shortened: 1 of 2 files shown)


kotlin-react-dom/src/main/generated/react/dom/PointerType.kt
```diff
@@ -0,0 +1,14 @@
...
+
+@Suppress("NAME_CONTAINS_ILLEGAL_CHARS")
+// language=JavaScript
+@JsName("""({mouse: 'mouse', pen: 'pen', touch: 'touch'})""")
+external enum class PointerType {
+    mouse,
+    pen,
+    touch,
+
+    ;
```
sgrishchenko, line 13: Why do you use semicolon, if there is trailing coma?

turansky: It's delimiter for constants and members. New members will not produce delimiters

sgrishchenko (approved):

Leonya:
Released in pre.232

PR #640, opened 2021-08-20
React HTML unions [generated]

turansky:
`InputType` union isn't declared in React types! :)

cc @sgrishchenko @aerialist7


Leonya:
Released in pre.234

PR #647, opened 2021-08-22
`RProps` is removed from react-router-dom

sgrishchenko:
cc @turansky @aerialist7

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
@@ -13,7 +13,7 @@ fun RBuilder.hashRouter(
     basename: String = "",
     hashType: HashType = HashType.slash,
     getUserConfirmation: GetUserConfirmation? = null,
-    handler: RHandler<PropsWithChildren>,
+    handler: Render,
 ) {
     HashRouter {
         attrs {
@@ -31,7 +31,7 @@ fun RBuilder.browserRouter(
     getUserConfirmation: GetUserConfirmation? = null,
     forceRefresh: Boolean = false,
     keyLength: Int = 6,
-    handler: RHandler<PropsWithChildren>,
+    handler: Render,
 ) {
     BrowserRouter {
         attrs {
@@ -46,7 +46,7 @@ fun RBuilder.browserRouter(
 }
 
 fun RBuilder.switch(
-    handler: RHandler<PropsWithChildren>,
+    handler: Render,
 ) {
     Switch(handler)
 }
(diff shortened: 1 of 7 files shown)


kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
```diff
@@ -13,7 +13,7 @@ fun RBuilder.hashRouter(
     basename: String = "",
     hashType: HashType = HashType.slash,
     getUserConfirmation: GetUserConfirmation? = null,
-    handler: RHandler<RProps>,
+    handler: RHandler<PropsWithChildren>,
```
turansky, line 16: `Props`? +4 else handlers

turansky (changes requested):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
turansky: Or even `RHandler<*>`?

sgrishchenko: I think it is correct change, because later it will be just `PropsWithChildren.() -> Unit`.
Of course, `hashRouter` and other routers have children, [see docs](https://reactrouter.com/web/api/Router):
```js
ReactDOM.render(
  <Router history={history}>
    <App />
  </Router>,
  node
);
```

sgrishchenko: Oh, think I've got you idea, we don't need `attrs` builder at all here, I will perform corrections

turansky: `attrs` for `key` and `ref` can be used

sgrishchenko: fixed

sgrishchenko: Yea, you are write about `key` and `ref`, but `RHandler<Props>` is not suitable here because we should be able to setup children in builder

sgrishchenko: Shall I just revert my last commit?

turansky: > RHandler<Props> is not suitable here because we should be able to setup children in builder

Children will work with `Props` :)

turansky (approved):

Leonya:
Released in pre.235

PR #666, opened 2021-08-26
Configure repositories in settings

turansky:
[Related issue](https://youtrack.jetbrains.com/issue/KT-48410)

cc @sgrishchenko @aerialist7


Leonya:
Released in pre.239

PR #674, opened 2021-08-29
Use `Record` for `Route` `Params` to prevent wrong typings

sgrishchenko:
### Problem
With existing implementation I can declare something like this:
```kotlin
external interface MyRouteParams : Props {
    val first: Int
    val second: Int
}

val MyComponent = fc<Props> {
    val params = useParams<MyRouteParams>() ?: return@fc

    println(params.first + params.second) // => for path /:first/:second and url /1/2 it will be string "12"
}
```
### Solution
Prevent opportunity of creation `Params` type with non-string types of values. Use just `Record<String. String>` or `Params` type alias.

### Migration 
Before:
```kotlin
external interface SomeProps : Props {
    val someId: String
}

val someId = useParams<SomeProps>()?.someId
```
After:
```kotlin
val someId = useParams()["someId"]
```
Also you can declare constant with you param and reuse it in you path:
```kotlin
val SOME_ID = "someId"

route("some/path/:$SOME_ID", component = Component)

val someId = useParams()[SOME_ID]
```  

cc @turansky @aerialist7

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
@@ -57,7 +57,7 @@ fun RBuilder.route(
     exact: Boolean = false,
     strict: Boolean = false,
 ) {
-    child<RouteProps<Props>, Route<Props>> {
+    Route {
         attrs {
             this.path = path
             this.exact = exact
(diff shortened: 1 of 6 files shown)


kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
```diff
@@ -81,34 +81,18 @@ fun RBuilder.route(
...
-        }
-    }
-}
-
 fun RBuilder.route(
     vararg path: String,
     exact: Boolean = false,
     strict: Boolean = false,
-    render: Render,
+    render: RBuilder.(RouteResultProps) -> Unit,
```
turansky, line 88: RHandler?

turansky (changes requested):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.kt
sgrishchenko: `typealias RHandler<P> = RElementBuilder<P>.() -> Unit`
`RHandler` does not do that is needed here

turansky (approved):

sgrishchenko:
@Leonya, could you make release to check these changes?

Leonya:
Released in pre.239

PR #693, opened 2021-08-31
React Router DOM: `Props` improvements

aerialist7:
cc @turansky @sgrishchenko

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/Route.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/Route.kt
@@ -3,23 +3,22 @@
 
 package react.router.dom
 
-import react.ComponentClass
-import react.ComponentType
-import react.Props
-import react.ReactElement
+import react.*
 
-external interface RouteProps : Props {
+external interface RouteProps : PropsWithChildren {
+    var component: ComponentType<*>
+    var render: (props: RouteComponentProps) -> ReactElement?
     var path: Array<out String>
     var exact: Boolean
+    var sensitive: Boolean
     var strict: Boolean
-    var component: ComponentType<*>
-    var render: (props: RouteResultProps) -> ReactElement?
 }
 
-external interface RouteResultProps : Props {
+external interface RouteComponentProps : Props {
     var history: History
     var location: Location
     var match: Match
+    var staticContext: StaticContext
 }
 
 external val Route: ComponentClass<RouteProps>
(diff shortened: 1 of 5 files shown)


kotlin-react-router-dom/src/main/kotlin/react/router/dom/Route.kt
```diff
@@ -3,23 +3,22 @@
...
 package react.router.dom
 
-import react.ComponentClass
-import react.ComponentType
-import react.Props
-import react.ReactElement
+import react.*
 
-external interface RouteProps : Props {
+external interface RouteProps : PropsWithChildren {
```
turansky, line 8: No `children` detected

turansky (changes requested):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/Route.kt
aerialist7: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e12cc2aedf5ee2ff668a9b295231d0c32af35285/types/react-router/index.d.ts#L86

turansky (approved):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/Route.kt
sgrishchenko: I am agree with @aerialist7. Children is most modern API and it is recommended since hooks promotion

sgrishchenko: https://reactrouter.com/web/api/Route/route-render-methods
> The recommended method of rendering something with a <Route> is to use children elements

kotlin-react-router-dom/src/main/kotlin/react/router/dom/Route.kt
```diff
@@ -3,9 +3,12 @@
 
 package react.router.dom
 
-import react.*
+import react.ComponentClass
+import react.ComponentType
+import react.Props
+import react.ReactElement
 
-external interface RouteProps : PropsWithChildren {
```
sgrishchenko: Nooo... You was right 🙂

aerialist7: Seems that in nearest future.
You can talk about it with Victor

sgrishchenko: https://reactrouter.com/web/api/Route/route-render-methods
> The recommended method of rendering something with a <Route> is to use children elements

Leonya:
Released in pre.241

PR #695, opened 2021-08-31
Router: "raw" prefix is removed for low-level wrappers

sgrishchenko:
@turansky @aerialist7

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/useRouteMatch.ext.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/useRouteMatch.ext.kt
@@ -8,10 +8,6 @@ fun useRouteMatch(
     strict: Boolean = false,
     sensitive: Boolean = false,
 ): Match? {
-    if (path.isEmpty()) {
-        return useRouteMatch()
-    }
-
     val options: MatchOptions = jso {
         this.path = path
         this.exact = exact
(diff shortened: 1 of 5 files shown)


kotlin-react-router-dom/src/main/kotlin/react/router/dom/useRouteMatch.kt
```diff
@@ -3,7 +3,6 @@
 
 package react.router.dom
 
-@JsName("useRouteMatch")
-external fun rawUseRouteMatch(
+external fun useRouteMatch(
     options: MatchOptions?,
```
turansky, line 7: definedExternally?

sgrishchenko: yea, thanks, fixed

kotlin-react-router-dom/src/main/kotlin/react/router/dom/useRouteMatch.ext.kt
```diff
@@ -9,7 +9,7 @@ fun useRouteMatch(
     sensitive: Boolean = false,
 ): Match? {
     if (path.isEmpty()) {
```
turansky: Never called :)

turansky (changes requested):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/useRouteMatch.ext.kt
sgrishchenko: fixed

turansky (approved):

Leonya:
Released in pre.241

PR #697, opened 2021-09-01
Children are provided for RouteProps, render prop is deprecated

sgrishchenko:
@turansky @aerialist7

### DIFF
--- a/kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.legacy.kt
+++ b/kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.legacy.kt
@@ -105,7 +105,7 @@ fun RBuilder.route(
     vararg path: String,
     exact: Boolean = false,
     strict: Boolean = false,
-    render: RBuilder.() -> Unit,
+    render: Render,
 ) {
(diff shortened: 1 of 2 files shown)


kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.legacy.kt
```diff
@@ -105,15 +105,16 @@ fun RBuilder.route(
     vararg path: String,
     exact: Boolean = false,
     strict: Boolean = false,
-    render: RBuilder.(RouteComponentProps) -> Unit,
+    render: RBuilder.() -> Unit,
```
turansky, line 108: `Render`?

turansky (changes requested):

turansky (approved):

kotlin-react-router-dom/src/main/kotlin/react/router/dom/routingDsl.legacy.kt
sgrishchenko: fixed

Leonya:
Released in pre.241

PR #706, opened 2021-09-04
Fix `react-redux` `Provider` declaration

turansky:
cc @sgrishchenko @aerialist7

### DIFF
--- a/kotlin-react-redux/src/main/kotlin/react/redux/Imports.kt
+++ b/kotlin-react-redux/src/main/kotlin/react/redux/Imports.kt
@@ -6,9 +6,7 @@ package react.redux
 import react.*
 import redux.Store
 
-external class Provider : Component<ProviderProps, State> {
-    override fun render(): ReactElement?
-}
+external val Provider: ComponentClass<ProviderProps>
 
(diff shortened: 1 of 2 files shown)


kotlin-react-redux/src/main/kotlin/react/redux/Imports.kt
```diff
@@ -6,9 +6,7 @@ package react.redux
 import react.*
 import redux.Store
 
-external class Provider : Component<ProviderProps, State> {
-    override fun render(): ReactElement?
-}
+external val Provider: ComponentClass<ProviderProps>
```
sgrishchenko, line 9: This is not `ComponentClass`, it is function component in redux source code, so maybe it is better to declare it as `ComponentType`?

sgrishchenko (changes requested):

kotlin-react-redux/src/main/kotlin/react/redux/Imports.kt
turansky: Fixed

Leonya:
Released in pre.242

PR #707, opened 2021-09-05
Add `react-popper` wrappers

turansky:
cc @sgrishchenko @aerialist7


kotlin-react-popper/src/main/kotlin/popper/PopperOptions.kt
```diff
@@ -0,0 +1,10 @@
+package popper
+
+import kotlinext.js.ReadonlyArray
+
+external interface PopperOptions<Modifiers> {
+    var placement: Placement
+    var modifiers: ReadonlyArray<Modifier<Modifiers>> // $Shape
+    var strategy: PositioningStrategy
+    var onFirstUpdate: (State) -> Unit // $Shape
```
sgrishchenko, line 9: type alias `PopperState`?

turansky: It's `State` from `popper` package

turansky: `PopperState` looks better

Leonya:
orgjetbrainskotlin-2002

PR #712, opened 2021-09-07
Fix add classname in stylesheet and add some tests

tretikoff:
Fix bug with classname adding in stylesheets and add tests for non-tested functionality found while debugging


Leonya:
Released in pre.243

PR #720, opened 2021-09-12
CSSType initial interfaces [generated]

turansky:
cc @sgrishchenko @aerialist7


kotlin-csstype/src/main/generated/csstype/StandardProperties.kt
```diff
@@ -0,0 +1,5 @@
+// Automatically generated - do not modify!
+
+package csstype
+
+external interface StandardProperties : StandardLonghandProperties
```
sgrishchenko, line 5: Why are you skipped `StandardShorthandProperties`?

sgrishchenko (approved):

kotlin-csstype/src/main/generated/csstype/StandardProperties.kt
turansky: Longhand properties first

Leonya:
Released in pre.244

PR #725, opened 2021-09-13
CSSType. KDoc initial fix [generated]

turansky:


sgrishchenko (approved):

kotlin-csstype/src/main/generated/csstype/StandardShorthandProperties.kt
```diff
@@ -10,11 +10,11 @@ external interface StandardShorthandProperties {
      *
      * **Initial value**: There is no practical initial value for it.
      *
-     * | Chrome | Firefox | Safari  |  Edge  | IE  |
-     * | :----: | :-----: | :-----: | :----: | :-: |
-     * | **37** | **27**  | **9.1** | **79** | No  |
+     *    | Chrome | Firefox | Safari  |  Edge  | IE  |
+     *    | :----: | :-----: | :-----: | :----: | :-: |
+     *    |   37   |   27    |   9.1   |   79   | No  |
```
sgrishchenko, line 15: It is strange that we have to remove mardown syntax, according to KDoc it should work normally, is there some problems with rendering of mardown inside table?

Leonya:
Released in pre.244

PR #732, opened 2021-09-14
Remove unused/invalid `author` property

turansky:
cc @aerialist7

### DIFF
--- a/kotlin-css/gradle.properties
+++ b/kotlin-css/gradle.properties
@@ -1,2 +1,1 @@
 description=Platform-agnostic CSS primitives (common module)
-author=Leonid Khachaturov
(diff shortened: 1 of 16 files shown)


Leonya:
It's not unused or invalid, it's a must for Maven Central publication...

Leonya:
You are right, it is unused, but it should be. Instead of deleting this property, the values should be propagated to POM files, I think.

turansky:
Default `developers` section, like in [typical JetBrains libraries](https://search.maven.org/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-core/1.5.2/pom) looks more logically. 
Also in most cases we have no fixed developers list

Leonya:
Released in pre.245

PR #744, opened 2021-09-15
CSSType. Missed `AutoLengthProperty` [generated]

turansky:


sgrishchenko:
How do you plan to use such `AutoLengthProperty`? Will it be something like that:
```kotlin
css {
  width =  AutoLengthProperty.auto
}
```
If so, it looks quite awkward. In typescript I can do such thing:
```ts
import {Properties} from 'csstype'

const value = "auto"

const css: Properties = {
    width: value,
    height: value,
    alignSelf: value,
}
```
See: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwApQmAprY6DOBfOAM1RDgHIBjHHGATwzIChGKIA7GuANwEMAbAK7o4AXjgAiHgJgRxzVh3hUcALjgo0mGNhyjEjOIbgB3YABMYACzW9B6ADQGjl9MADmlmDf5DHRuPzubADK6HyE3naOeEA

How do you plan to support this implemented in Kotlin?
I think good reference here is `Unit` object, that common for every usage, I propose to implement it like this
```kotlin
JsName("{auto: 'auto'}")
external interface Width {
  val auto: Auto
}

JsName("{auto: 'auto'}")
external interface Height {
  val auto: Auto
}

JsName("{auto: 'auto'}")
external interface AlignSelf {
  val auto: Auto
}

object Auto: Width, Height, AlignSelf // and other usages

val value = Auto

css {
  width =  value
  height = value
  alignSelf = value
}

// or just

css {
  width =  Width.auto
}
```

Leonya:
Released in pre.246

PR #771, opened 2021-09-18
CSSType. `HtmlAttributes` - enum [generated]

turansky:


kotlin-csstype/src/main/generated/csstype/HtmlAttributes.kt
```diff
@@ -1,229 +1,238 @@
...
+    target,
+    text,
+    title,
+    topmargin,
+    translate,
+    truespeed,
+    type,
+    usemap,
+    valign,
+    __value__,
```
sgrishchenko, line 226: @turansky I think we should use special syntax for it `value`

Leonya:
Released in pre.247

PR #811, opened 2021-09-25
CSSType. Advanced `PseudosRuleBuilder` [generated]

turansky:


kotlin-csstype/src/main/generated/csstype/AdvancedPseudosRuleBuilder.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    inline fun hostContext(
+        selector: String,
+        block: T.() -> Unit,
+    ) {
+        ":host-context($selector)"(block)
+    }
+
+    inline fun __is__(
```
sgrishchenko, line 66: maybe 
```
`is`
```

sgrishchenko (changes requested):

kotlin-csstype/src/main/generated/csstype/AdvancedPseudosRuleBuilder.kt
turansky: It's common generator rule (separate PR in any case)

Leonya:
Released in pre.249

PR #816, opened 2021-09-26
CSSType. Remove `@viewport` types [generated]

turansky:
[`@viewport` deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/@viewport)


Leonya:
Released in pre.249

PR #831, opened 2021-10-01
New signature of `createElement` is introduced

sgrishchenko:
New signature of `createElement` is introduced, deprecation for old implementation is added.

cc @turansky @aerialist7

### DIFF
--- a/kotlin-react/src/main/kotlin/react/RBuilder.kt
+++ b/kotlin-react/src/main/kotlin/react/RBuilder.kt
@@ -71,7 +71,7 @@ interface RBuilder {
         handler: RBuilder.(T) -> Unit,
     ) {
         child(this) {
-            attrs.children = { value -> buildElements { handler(value) } }
+            attrs.children = { value -> createElement { handler(value) } }
         }
     }
 
@@ -140,6 +140,7 @@ open class RBuilderImpl : RBuilder {
 
 open class RBuilderMultiple : RBuilderImpl()
 
+@Deprecated(message = "Unsafe API, use createElement")
 fun <T : RBuilder> buildElements(builder: T, handler: T.() -> Unit): dynamic {
     val nodes = builder.apply(handler).childList
     return when (nodes.size) {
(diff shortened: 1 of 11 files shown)


kotlin-react/src/main/kotlin/react/ReactElement.ext.kt
```diff
@@ -0,0 +1,19 @@
+package react
+
+import kotlinext.js.jso
+
+inline fun createElement(block: Render): ReactElement? {
```
turansky, line 5: inline?

sgrishchenko: copy-paste :stuck_out_tongue_winking_eye:, fixed

kotlin-react/src/main/kotlin/react/ReactElement.ext.kt
```diff
@@ -0,0 +1,19 @@
+package react
+
+import kotlinext.js.jso
+
+fun createElement(block: Render): ReactElement? {
```
turansky, line 5: Discussion required

kotlin-react/src/main/kotlin/react/ReactElement.ext.kt
```diff
@@ -0,0 +1,19 @@
...
+    val nodes = RBuilder().apply(block).childList
+
+    return when {
+        nodes.size == 0 -> null
+        nodes.size == 1 && isValidElement(nodes.first()) -> nodes.first().unsafeCast<ReactElement>()
+        else -> createElement(Fragment, jso(), *nodes.toTypedArray())
```
turansky, line 11: Named parameter?

Leonya:
Released in pre.252

kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -140,6 +140,7 @@ open class RBuilderImpl : RBuilder {
 
 open class RBuilderMultiple : RBuilderImpl()
 
+@Deprecated(message = "Unsafe API, use createElement")
```
Leonya, line 143: How am I supposed to use it if `createElement` doesn't have the `builder` parameter?

PR #835, opened 2021-10-03
React. Use `ReadonlyArray`

turansky:


kotlin-react/src/main/kotlin/react/useEffect.kt
```diff
@@ -1,6 +1,8 @@
 package react
 
-typealias RDependenciesArray = Array<out dynamic>
+import kotlinext.js.ReadonlyArray
+
+typealias RDependenciesArray = ReadonlyArray<dynamic>
```
sgrishchenko, line 5: this type is used not only in effect hooks, maybe it is good opportunity to move it in some common place?

sgrishchenko (approved):

kotlin-react/src/main/kotlin/react/useEffect.kt
turansky: It has sense to inline it in separate PR

Leonya:
Released in pre.252

PR #837, opened 2021-10-03
`Dependencies` type alias is introduced, `RDependenciesArray` is deprecated

sgrishchenko:
cc @turansky @aerialist7

### DIFF
--- a/kotlin-react/src/main/kotlin/react/Dependencies.kt
+++ b/kotlin-react/src/main/kotlin/react/Dependencies.kt
@@ -4,8 +4,3 @@ import kotlinext.js.ReadonlyArray
 
 typealias Dependencies = ReadonlyArray<dynamic>
 
-@Deprecated(
-    message = "Legacy API",
-    replaceWith = ReplaceWith("Dependencies", "react.Dependencies")
-)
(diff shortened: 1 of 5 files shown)


kotlin-react/src/main/kotlin/react/Dependencies.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import kotlinext.js.ReadonlyArray
+
+typealias Dependencies = ReadonlyArray<dynamic>
+
+@Deprecated(
+    message = "Legacy API",
+    replaceWith = ReplaceWith("Dependencies", "react.Dependencies")
+)
+typealias RDependenciesArray = Dependencies
```
turansky, line 11: In separate file please

turansky (changes requested):

kotlin-react/src/main/kotlin/react/Dependencies.kt
sgrishchenko: fixed

turansky (approved):

Leonya:
Released in pre.252

PR #921, opened 2021-11-02
Regular cleanup of deprecated APIs

turansky:
cc @aerialist7 

Previous - #662


kotlin-react/src/main/kotlin/react/RBuilder.kt
```diff
@@ -133,22 +106,6 @@ open class RBuilderImpl : RBuilder {
 
 open class RBuilderMultiple : RBuilderImpl()
 
-@Deprecated(message = "Unsafe API, use createElement")
-fun <T : RBuilder> buildElements(builder: T, handler: T.() -> Unit): dynamic {
```
Leonya, line 137: Still using this one.

Leonya:
renderEach can actually be removed.

kotlin-react/src/main/kotlin/react/RBuilder.kt
turansky: Restored

turansky:
`renderEach` removed

PR #930, opened 2021-11-07
CSSType. `Content` factory methods

turansky:


kotlin-csstype/src/main/kotlin/csstype/Content.ext.kt
```diff
@@ -0,0 +1,15 @@
...
+package csstype
+
+inline fun Content(
+    value: String,
+): Content =
+    "'$value'".unsafeCast<Content>()
+
+inline fun Content(
+    value: Int,
+): Content =
```
sgrishchenko, line 14: why is only overload for `Int` added? I think `Content(String)` is enough. 
Moreover I think we don't need to make assumptions how developers will use this method, I think all user of css-in-js solutions are used to this syntax `Content("'value'")`

sgrishchenko (changes requested):

sgrishchenko:
more discussion is required

Issue #711, opened 2021-09-07
Why are the router extensions deprecated?


joffrey-bion:
Since I upgraded to latest versions of Kotlin wrappers, my usages of router extensions are marked as deprecated.
The `ReplaceWith` is broken so I had to manually replace the function calls.

What is the reason for the deprecation? Is there an alternative to using the raw component directly?

It really looks awkward now. I used to have this:

```kotlin
fun RBuilder.application() = hashRouter {
    errorDialog()
    switch {
        route(Route.GAME_BROWSER.path) { gameBrowser() }
        route(Route.GAME.path) { gameScene() }
        route(Route.LOBBY.path) { lobby() }
        route(Route.HOME.path, exact = true) { home() }
        redirect(from = "*", to = "/")
    }
}
```
And now I have to replace with:
```kotlin
fun RBuilder.application() = HashRouter {
    errorDialog()
    Switch {
        Route {
            attrs.path = arrayOf(Route.GAME_BROWSER.path)
            gameBrowser()
        }
        Route {
            attrs.path = arrayOf(Route.GAME.path)
            gameScene()
        }
        Route {
            attrs.path = arrayOf(Route.LOBBY.path)
            lobby()
        }
        Route {
            attrs.path = arrayOf(Route.HOME.path)
            home()
        }
        Redirect {
            attrs {
                from = "*"
                to = "/"
            }
        }
    }
}
```

sgrishchenko:
@joffrey-bion thank you for your feedback. There was few reasons for deprecation router extensions:
1) Name conversions: in React usually only DOM elements starts with small letter
1) Default values duplication: route flags (e.g. `exact`) had default values to make them optional, and that values were copy-pasted from React Router API. It was not good idea because that defaults can be changed and it would lead to unexpected behavior of `kotlin-react-router`
1) Deprecated extensions did nothing: they didn't add any useful behavior, they just converted block signature to function call so there were few reasons to duplicate almost all declarations only for that purpose.

P. S. Sorry for broaken `ReplaceWith`, it seems we should not have added it at all

joffrey-bion:
Thanks a lot for the detailed reasons! These are totally understandble.

I wish there was a shorter way to set the props of class components, but I guess that will do for the moment.

> Sorry for broaken ReplaceWith, it seems we should not have added it at all

No worries, I'm not even sure that it's possible to have a correct `ReplaceWith` for this.

PR #938, opened 2021-11-10
Lightweight builder (without `attrs`)

turansky:
Related issues:
#711, #614

image


kotlin-react/src/main/kotlin/react/experimental/createElement.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import kotlinext.js.ReadonlyArray
+import react.*
+
+fun createElement(
+    block: Render,
+): ReactElement? {
+    val children: ReadonlyArray<ReactNode>? = ChildrenBuilder()
+        .apply(block)
+        .asDynamic()
```
sgrishchenko, line 11: it seems `asDynamic` can be omitted

sgrishchenko (changes requested):

kotlin-react/src/main/kotlin/react/experimental/createElement.kt
turansky: No `children` property in `ChildrenBuilder`

sgrishchenko: If so, what is it?
```
private inline var ChildrenBuilder.children: ReadonlyArray<ReactNode>?
```

sgrishchenko (changes requested):

kotlin-react/src/main/kotlin/react/experimental/createElement.kt
turansky: Fixed

sgrishchenko (approved):

kotlin-react-dom/src/main/kotlin/react/dom/experimental/render.kt
```diff
@@ -0,0 +1,16 @@
+package react.dom.experimental
+
+import org.w3c.dom.Element
+import react.dom.render
+import react.experimental.Render
+import react.experimental.createElement
+
+fun render(
+    container: Element,
```
sgrishchenko, line 9: `Element?` is better here, because e.g. `document.getElementById` returns `Element?`

kotlin-react/src/main/kotlin/react/experimental/createElement.kt
```diff
@@ -0,0 +1,19 @@
...
+fun createElement(
+    block: Render,
+): ReactElement? {
+    val children: ReadonlyArray<ReactNode>? = ChildrenBuilder()
+        .apply(block)
+        .children
+
+    return when {
+        children == null -> null
+        children.size == 0 -> null
```
sgrishchenko, line 15: Can be replaced with `children.isEmpty()` (IDEA suggestion)

PR #940, opened 2021-11-10
CssBuilder structural equality

tretikoff:


kotlin-styled-next/src/main/kotlin/styled/StructEffects.kt
```diff
@@ -0,0 +1,54 @@
...
+                prevResultRef.current = MemoizedResult(args, result, cleanups)
+            }
+        }
+    }
+}
+
+private fun <T> MemoizedResult<T>?.argsEqual(args: Array<out dynamic>): Boolean {
+    if (this == null) return false
+    if (this.args.size != args.size) {
+        if (isDevelopment()) console.warn("Calling useEffect with different length dependencies: ${this.args.joinToString()} and ${args.joinToString()}")
```
Leonya, line 45: "length" -> "number of"

kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
```diff
@@ -17,7 +17,7 @@ class Import(private val url: String, private val types: List<String> = listOf()
  *
  * [name] is the stylesheet identifier. CSS class names are generated using a combination of [name] and property name.
  *
- * If [isStatic] property is set to true, a static, human-readable class selector will be created for every property,
+ * If [isStatic] prperty is set to true, a static, human-readable class selector will be created for every property,
```
Leonya, line 20: Typo here

Leonya (changes requested):
Please reformat all modified files.

kotlin-css/src/commonMain/kotlin/kotlinx/css/CssBuilder.kt
```diff
@@ -116,6 +119,10 @@ interface CssBuilder : StyledElement, RuleContainer {
         val selectorString = selector.joinToString { "> $it" }
         return selectorString(block)
     }
+    fun child(vararg selector: String, css: CssBuilder): Rule {
```
Leonya, line 122: If this is needed, then what about all other methods?

Leonya (changes requested):

kotlin-css/src/commonMain/kotlin/kotlinx/css/CssBuilder.kt
tretikoff: It is only needed for `compareTo` method, which uses other rule (already converted to CssBuilder because we don't store lambdas)

Leonya (approved):

PR #952, opened 2021-11-18
React Router. Kotlinish `useSearchParams` [generated]

turansky:


kotlin-react-router-dom/src/main/generated/react/router/dom/URLSearchParamsInit.kt
```diff
@@ -2,4 +2,4 @@
 
 package react.router.dom
 
-typealias URLSearchParamsInit = String // string | ParamKeyValuePair[] | Record<string, string | string[]> | URLSearchParams
\ No newline at end of file
+typealias URLSearchParamsInit = kotlinext.js.Record<String, String> // string | ParamKeyValuePair[] | Record<string, string | string[]> | URLSearchParams
```
sgrishchenko, line 5: How can I use this signature for arrays? Example: `array=one&array=two&array=three'`

PR #961, opened 2021-11-20
Popper. Fix `Modifier` type parameters

turansky:


JerCol:
Dear, I don't know if this question has its place here, but I do not know how I define a possible options argument in the usePopper method (see below). Can you provide an example? At the moment I am not capable to use the kotlin-react-popper library.
 
react.popper popper.kjsm public external fun <Modifiers> usePopper(
    referenceElement: Element?,
    popperElement: HTMLElement?,
    options: UsePopperOptions<Modifiers>
): PopperInstance

turansky:
```kotlin
val popper = usePopper(
    referenceElement  = referenceElement,
    popperElement = popperElement,
    options = jso {
        // ...
    }
)
```

PR #963, opened 2021-11-21
Popper. Base types [generated]

turansky:
Related - #956


kotlin-popper/src/main/generated/popper/core/Padding.kt
```diff
@@ -0,0 +1,5 @@
+// Automatically generated - do not modify!
+
+package popper.core
+
+typealias Padding = SideObject // number | Partial<SideObject>
```
sgrishchenko, line 5: it seems it is too strict limitations because original type is `Partial<SideObject>`

sgrishchenko (approved):

PR #964, opened 2021-11-21
`useImperativeHandle` hook is refactored

sgrishchenko:
cc @turansky @aerialist7

### DIFF
--- a/examples/src/main/kotlin/example/ForwardingRefs.kt
+++ b/examples/src/main/kotlin/example/ForwardingRefs.kt
@@ -1,0 +1,41 @@
+package example
+
+import org.w3c.dom.HTMLInputElement
+import react.*
+import react.dom.html.ReactHTML.input
+
+external interface Focusable {
+    fun focus()
+}
+
+val FancyInput = forwardRef<Props> { _, forwardedRef ->
+    val inputRef = useRef<HTMLInputElement>()
+
+    useImperativeHandle(forwardedRef, inputRef) {
+        object : Focusable {
+            override fun focus() {
+                inputRef.current?.focus()
+            }
+        }
+    }
+
+    input {
+        attrs {
+            ref = inputRef
+        }
+    }
+}
+
+val App = fc<Props> {
+    val fancyInputRef = useRef<Focusable>()
+
+    useEffectOnce {
+        fancyInputRef.current?.focus()
+    }
+
+    FancyInput {
+        attrs {
+            ref = fancyInputRef
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


examples/src/main/kotlin/example/ForwardingRefs.kt
```diff
@@ -0,0 +1,41 @@
...
+    useImperativeHandle(forwardedRef, inputRef) {
+        object : Focusable {
+            override fun focus() {
+                inputRef.current?.focus()
+            }
+        }
+    }
+
+    input {
+        attrs {
```
turansky, line 23: `attrs` can be removed
`react.dom.html.ref` - Strict ref

examples/src/main/kotlin/example/ForwardingRefs.kt
```diff
@@ -0,0 +1,41 @@
...
+
+external interface Focusable {
+    fun focus()
+}
+
+val FancyInput = forwardRef<Props> { _, forwardedRef ->
+    val inputRef = useRef<HTMLInputElement>()
+
+    useImperativeHandle(forwardedRef, inputRef) {
+        object : Focusable {
```
turansky, line 15: `jso`?
Do you really need additional class?

turansky (changes requested):

examples/src/main/kotlin/example/ForwardingRefs.kt
```diff
@@ -0,0 +1,41 @@
...
+val App = fc<Props> {
+    val fancyInputRef = useRef<Focusable>()
+
+    useEffectOnce {
+        fancyInputRef.current?.focus()
+    }
+
+    FancyInput {
+        attrs {
+            ref = fancyInputRef
```
turansky, line 38: Request discussion about "forward ref component"

turansky (changes requested):

examples/src/main/kotlin/example/ForwardingRefs.kt
sgrishchenko: in `React` ref is regular prop, I don't know why it was moved to `RElementBuilder`

examples/src/main/kotlin/example/ForwardingRefs.kt
sgrishchenko: yes I do, I want to strictly implement external contract, `jso` is unsafe for it

examples/src/main/kotlin/example/ForwardingRefs.kt
sgrishchenko: All in all, I think we can merge this PR

PR #969, opened 2021-11-22
React. Sealed component type interfaces

turansky:
cc @aerialist7


Leonya:
Eh... How's one supposed to implement those interfaces if they're sealed?

turansky:
We didn't find use cases for custom implementation

cc @sgrishchenko

sgrishchenko (approved):

Leonya:
I have several `ComponentClass` implementations. Used to be `RClass` before. Should it be something else?

turansky:
`ComponentClass`  signature restored

PR #1033, opened 2021-12-06
`SuspenseProps` extension is moved to separate file

sgrishchenko:
cc @turansky

### DIFF
--- a/kotlin-react/src/main/kotlin/react/Suspense.ext.kt
+++ b/kotlin-react/src/main/kotlin/react/Suspense.ext.kt
@@ -4,5 +4,5 @@ import react.createElement
 
 // 16.6+
 fun SuspenseProps.fallback(handler: Render) {
-    asDynamic().fallback = createElement(handler)
+    fallback = createElement(handler)
 }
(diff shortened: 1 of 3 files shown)


kotlin-react/src/main/kotlin/react/Suspense.ext.kt
```diff
@@ -0,0 +1,8 @@
+import react.Render
+import react.SuspenseProps
+import react.createElement
+
+// 16.6+
+fun SuspenseProps.fallback(handler: Render) {
+    asDynamic().fallback = createElement(handler)
```
turansky, line 7: Redundant `asDynamic`?

turansky (changes requested):

kotlin-react/src/main/kotlin/react/Suspense.ext.kt
sgrishchenko: `SuspenseProps` are decalared next way now:
```
external interface SuspenseProps : PropsWithChildren
```
Shell I declare it this way:
```
external interface SuspenseProps : PropsWithChildren {
    var fallback: ReactNode?
}
```

turansky: [Yes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b8e894eaa4ce70830a40a6bcc3c12392eb867989/types/react/index.d.ts#L400)

sgrishchenko: fixed

turansky (approved):

PR #1042, opened 2021-12-10
[ring-ui] Add Table wrapper

Neitex:
The problem is that Kotlin Set does not map to JS Set, which Table (and some other) components use. But adding a whole JS Set wrapper seems to be kind of overkill. So, this PR is mainly made for some discussion about what to do. However, Table wrapper in this PR is implemented using new JS Set wrapper,


turansky:
Issue required instead
Question is quite common

cc @sgrishchenko

Neitex:
Should I close this PR or leave it as is?

turansky:
It can be closed and added as "related" in issue

PR #1049, opened 2021-12-13
React. `useDeferredValue`

turansky:
[Typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/46d4de0aeadaa63c82c069bbf6a8049bb0cd9968/types/react/next.d.ts#L71)

### DIFF
--- a/kotlin-react-core/src/main/kotlin/react/useDeferredValue.kt
+++ b/kotlin-react-core/src/main/kotlin/react/useDeferredValue.kt
@@ -1,0 +1,14 @@
+@file:JsModule("react")
+@file:JsNonModule
+
+
+package react
+
+/**
+ * Available from 18.0
+ *
+ * @see <a href="https://reactjs.org/docs/concurrent-mode-reference.html#usedeferredvalue">Online Documentation</a>
+ */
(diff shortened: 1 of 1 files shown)


kotlin-react-core/src/main/kotlin/react/useDeferredValue.kt
```diff
@@ -0,0 +1,14 @@
...
+package react
+
+/**
+ * Available from 18.0
+ *
+ * @see <a href="https://reactjs.org/docs/concurrent-mode-reference.html#usedeferredvalue">Online Documentation</a>
+ */
+external fun <T> useDeferredValue(
+    value: T,
+): T
```
sgrishchenko, line 14: Could you provide SUSPENSE_CONFIG here?

sgrishchenko (changes requested):

sgrishchenko (approved):

kotlin-react-core/src/main/kotlin/react/useDeferredValue.kt
turansky: Not found in sources

PR #1050, opened 2021-12-13
React. `useSyncExternalStore`

turansky:
[Typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/46d4de0aeadaa63c82c069bbf6a8049bb0cd9968/types/react/next.d.ts#L135)

### DIFF
--- a/kotlin-react-core/src/main/kotlin/react/useSyncExternalStore.kt
+++ b/kotlin-react-core/src/main/kotlin/react/useSyncExternalStore.kt
@@ -1,0 +1,15 @@
+@file:JsModule("react")
+@file:JsNonModule
+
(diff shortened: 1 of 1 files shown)


sgrishchenko (approved):

kotlin-react-core/src/main/kotlin/react/useSyncExternalStore.kt
```diff
@@ -0,0 +1,15 @@
+@file:JsModule("react")
+@file:JsNonModule
+
+package react
+
+/**
+ * Available from 18.0
```
vlsi, line 7: from -> since

PR #1055, opened 2021-12-14
Ignore `yarn.lock`

turansky:


Leonya:
Why ignore?

turansky:
No difference expected (after PR)
It will work like in `1.6.0`

PR #1089, opened 2022-01-02
MUI. `SortDirection` [generated]

turansky:
cc @aerialist7

### DIFF
--- a/kotlin-mui/src/main/generated/mui/material/TableCell.ext.kt
+++ b/kotlin-mui/src/main/generated/mui/material/TableCell.ext.kt
@@ -4,6 +4,17 @@ package mui.material
 
 @Suppress("NAME_CONTAINS_ILLEGAL_CHARS")
 // language=JavaScript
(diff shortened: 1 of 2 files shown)


kotlin-mui/src/main/generated/mui/material/TableCell.ext.kt
```diff
@@ -2,6 +2,17 @@
 
 package mui.material
 
+@Suppress("NAME_CONTAINS_ILLEGAL_CHARS")
+// language=JavaScript
+@JsName("""(/*union*/{asc: 'asc', desc: 'desc', false: 'false'}/*union*/)""")
```
aerialist7, line 7: It is `false` (`bool`, not `string`) in original.
https://mui.com/api/table-cell/

Will it be changed in future?

aerialist7 (changes requested):

kotlin-mui/src/main/generated/mui/material/TableCell.ext.kt
aerialist7: Reproduced with ``` SortDirection.`false` ```

image

PR #1097, opened 2022-01-03
Examples Simplification

aerialist7:
cc @turansky


examples/src/main/kotlin/example/Product.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("NonNullableBooleanPropertyInExternalInterface")
```
turansky, line 1: Redundant?

turansky (changes requested):

examples/src/main/kotlin/example/Product.kt
aerialist7: No

PR #1119, opened 2022-01-06
Add ByteLegend to examples

blindpirate:
Kotlin/JS is awesome, I've been developing an MMORPG game with Kotlin/JS for over a year. I'd like to add it as a "real-world example" to the list.

https://bytelegend.com/

image


Leonya:
Nice one!

PR #1135, opened 2022-01-16
Formatting

aerialist7:
cc @turansky


turansky (changes requested):
[And here?](https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-wrappers-bom/README.md)

turansky (approved):

PR #1139, opened 2022-01-18
MUI `5.3.0` [generated]

turansky:
cc @aerialist7


kotlin-mui/src/main/generated/mui/material/styles/createTheme.kt
```diff
@@ -24,21 +24,7 @@ external interface ThemeOptions {
 }
 
 external interface Theme {
```
sgrishchenko, line 26: Where is inheritance from `BaseTheme` and `SystemTheme` 
https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/styles/createTheme.d.ts#L21
https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/styles/createTheme.d.ts#L37

sgrishchenko (changes requested):

kotlin-mui/src/main/generated/mui/material/styles/createTheme.kt
turansky: Fixed in #1161

Issue #1137, opened 2022-01-17
TypeScript wrapper


turansky:
Source package - [`typescript`](https://www.npmjs.com/package/typescript)
Type - Node.js library
Converter - [`types-kotlin`](https://github.com/karakum-team/types-kotlin)

PR #1154, opened 2022-01-28
TypeScript. Sealed interfaces [generated]

turansky:
Related - #1137


sgrishchenko (changes requested):
I don't like these changes. I can understand reasons to seal `Node` hierarchy, but not all interfaces in library. What the reason to do it? I don't recommend to merge it.

turansky:
Current expectation - 5-20 non-sealed interfaces (from 500).
But we will know them only when we will have more experience.

PR #1215, opened 2022-02-17
CSSType. Remove unsafe factory methods

turansky:


goncalossilva:
This also eliminates the single parameter Margin/Padding invocation. Is that expected?

turansky:
@goncalossilva Which call do you mean?

goncalossilva:
For example, `Margin(auto)` does not work. Only `Margin(auto, auto)`.

turansky:
`margin = auto` :)

@sgrishchenko

goncalossilva:
Oh. I didn't figure that one out! Thanks for sharing. :)

PR #1236, opened 2022-02-20
`WeakMap` and `WeakSet`

### LINKED ISSUE #1231: Add WeakSet and WeakMap wrappers
Add WeakSet class with constructors `WeakSet(iterable?)` and methods :`add()`, `delete()`, `has()`

Add WeakMap with constructors `WeakMap(iterable?)`, `WeakMap()` and methods : `delete()`, `get()`, `has()`, `set()`

turansky:
Fix #1231

### DIFF
--- a/kotlin-extensions/src/main/kotlin/kotlinx/js/WeakMap.kt
+++ b/kotlin-extensions/src/main/kotlin/kotlinx/js/WeakMap.kt
@@ -1,0 +1,10 @@
+package kotlinx.js
+
+external class WeakMap<in K : Any, out V>(
+    entries: ReadonlyArray<JsPair<K, V>>? = definedExternally,
+) {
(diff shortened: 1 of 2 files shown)


kotlin-extensions/src/main/kotlin/kotlinx/js/WeakMap.kt
```diff
@@ -0,0 +1,10 @@
+package kotlinx.js
+
+external class WeakMap<in K : Any, out V>(
```
sgrishchenko, line 3: It seems `V` parameter should be invariant because `WeakMap` is mutable

turansky: Fixed

PR #1270, opened 2022-02-26
React. `ReactElement` type parameter

turansky:
cc @sgrishchenko @aerialist7 

@Leonya, we request release after this PR

### DIFF
--- a/examples/src/main/kotlin/example/ForwardingRefs.kt
+++ b/examples/src/main/kotlin/example/ForwardingRefs.kt
@@ -1,11 +1,12 @@
 package example
 
+import kotlinx.js.jso
 import org.w3c.dom.HTMLInputElement
 import react.*
 import react.dom.html.ReactHTML.input
 
 external interface Focusable {
-    fun focus()
+    var focus: () -> Unit
 }
 
 typealias FancyInputProps = PropsWithRef<Focusable>
@@ -14,8 +15,8 @@ val FancyInput = rawForwardRef<Focusable, FancyInputProps> { _, forwardedRef ->
     val inputRef = useRef<HTMLInputElement>()
 
     useImperativeHandle(forwardedRef, inputRef) {
-        object : Focusable {
-            override fun focus() {
+        jso {
+            focus = {
                 inputRef.current?.focus()
             }
         }
@@ -30,7 +31,7 @@ val ForwardingRefsApp = FC<Props> {
     val fancyInputRef = useRef<Focusable>()
 
     useEffectOnce {
-        fancyInputRef.current?.focus()
+        fancyInputRef.current?.focus?.invoke()
     }
 
     FancyInput {
--- a/kotlin-react-dom-legacy/src/main/kotlin/react/dom/createPortal.ext.kt
+++ b/kotlin-react-dom-legacy/src/main/kotlin/react/dom/createPortal.ext.kt
@@ -1,6 +1,7 @@
 package react.dom
 
 import org.w3c.dom.Element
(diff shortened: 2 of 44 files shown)


examples/src/main/kotlin/example/ForwardingRefs.kt
```diff
@@ -1,11 +1,12 @@
 package example
 
+import kotlinx.js.jso
 import org.w3c.dom.HTMLInputElement
 import react.*
 import react.dom.html.ReactHTML.input
 
 external interface Focusable {
-    fun focus()
+    var focus: () -> Unit
```
sgrishchenko, line 9: I don see the purpose for these changes, `Focusable` it is not props, it is immutable object, it is output data from components

sgrishchenko: @turansky if you want to discuss it, let's move these changes to separate PR

sgrishchenko (changes requested):

aerialist7 (approved):

kotlin-react-dom-legacy/src/main/kotlin/react/dom/createPortal.ext.kt
```diff
@@ -10,6 +11,6 @@ fun createPortal(
     handler: Render,
 ): ReactPortal =
     createPortal(
-        children = createElement(handler)!!,
+        children = createElement<Props>(handler)!!,
```
sgrishchenko, line 14: double exclamation mark is not needed here because `createPortal` should receive nullable `ReactNode`

kotlin-react-table/src/main/kotlin/react/table/Columns.kt
```diff
@@ -51,7 +51,7 @@ class ColumnBuilder<D : Any> {
         id: IdType<D>,
         header: String,
         accessor: D.() -> V,
-        cell: (CellProps<D, V>) -> ReactElement,
+        cell: (CellProps<D, V>) -> ReactElement<*>,
```
sgrishchenko, line 54: Are props really unknown here? Additional iterations are needed to determine concrete props?

examples/src/main/kotlin/example/ForwardingRefs.kt
sgrishchenko: `Focusable` is some kind of custom object, it is not declared is react types of in typescript typings in browser, what kind of symmetric are we talking about?

kotlin-react-dom-legacy/src/main/kotlin/react/dom/createPortal.ext.kt
sgrishchenko: Separate PR?

kotlin-react-table/src/main/kotlin/react/table/Columns.kt
turansky: Yes

kotlin-react-dom-legacy/src/main/kotlin/react/dom/createPortal.ext.kt
turansky: You are right - I will fix it in separate PR

examples/src/main/kotlin/example/ForwardingRefs.kt
sgrishchenko: It was decided to discuss these changes later.

sgrishchenko (approved):

PR #1292, opened 2022-03-04
JS. Timers

### LINKED ISSUE #1224: Add Timers wrappers
Add `setInterval`, `setTimeout`, `setImmediate` and their `clearThing` forms.

turansky:
Fix #1224

[Related issue](https://youtrack.jetbrains.com/issue/KT-46315)

cc @aerialist7 @stefanthaler

### DIFF
--- a/kotlin-extensions/src/main/kotlin/kotlinx/js/timers/interval.kt
+++ b/kotlin-extensions/src/main/kotlin/kotlinx/js/timers/interval.kt
@@ -1,0 +1,10 @@
+package kotlinx.js.timers
+
+external fun setInterval(
+    callback: () -> Unit,
+    ms: Double = definedExternally,
+): Timeout
+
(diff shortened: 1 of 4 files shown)


aerialist7 (approved):

kotlin-extensions/src/main/kotlin/kotlinx/js/timers/interval.kt
```diff
@@ -0,0 +1,10 @@
+package kotlinx.js.timers
+
+external fun setInterval(
+    callback: () -> Unit,
+    ms: Double = definedExternally,
```
sgrishchenko, line 5: Are you sure that we need `Double` here?
https://stackoverflow.com/questions/7286178/does-settimeout-in-javascript-accept-real-floating-point-delay-times

PR #1295, opened 2022-03-06
Emotion. Duplicate css prop to emotion project

sgrishchenko:
cc @turansky @aerialist7


turansky (changes requested):
I suppose, that we can avoid duplication

aerialist7 (approved):

turansky (approved):

PR #1299, opened 2022-03-06
Emotion. Keyframes helper is introduced

sgrishchenko:
cc @turansky @aerialist7

### DIFF
--- a/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
+++ b/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
@@ -32,6 +32,6 @@ interface KeyframesBuilder {
 }
 
 inline fun keyframes(
-    blocks: KeyframesBuilder.() -> Unit,
+    block: KeyframesBuilder.() -> Unit,
 ): AnimationName =
-    keyframes(jso(blocks).unsafeCast<Keyframes>())
+    keyframes(jso(block).unsafeCast<Keyframes>())
--- a/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
+++ b/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
@@ -34,4 +34,4 @@ interface KeyframesBuilder {
 inline fun keyframes(
     blocks: KeyframesBuilder.() -> Unit,
 ): AnimationName =
-    rawKeyframes(jso(blocks).unsafeCast<Keyframes>())
+    keyframes(jso(blocks).unsafeCast<Keyframes>())
--- a/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
+++ b/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
@@ -8,16 +8,30 @@ import csstype.*
 import kotlinx.js.Record
 import kotlinx.js.jso
 
+typealias Keyframes = Record<Percentage, Properties>
+
 @CssDsl
 interface KeyframesBuilder {
     inline operator fun Percentage.invoke(
         block: Properties.() -> Unit
     ) {
-        unsafeCast<Record<Percentage, Properties>>()[this] = jso(block)
+        unsafeCast<Keyframes>()[this] = jso(block)
+    }
+
+    inline fun from(
+        block: Properties.() -> Unit
+    ) {
+        0.pct(block)
+    }
+
+    inline fun to(
+        block: Properties.() -> Unit
+    ) {
+        100.pct(block)
     }
 }
 
 inline fun keyframes(
     blocks: KeyframesBuilder.() -> Unit,
 ): AnimationName =
-    rawKeyframes(jso(blocks).unsafeCast<Record<Percentage, Properties>>())
+    rawKeyframes(jso(blocks).unsafeCast<Keyframes>())
--- a/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
+++ b/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
@@ -10,7 +10,9 @@ import kotlinx.js.jso
 
 @CssDsl
 interface KeyframesBuilder {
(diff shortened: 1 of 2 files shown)


kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
```diff
@@ -0,0 +1,12 @@
...
+
+package emotion.css
+
+import csstype.AnimationName
+import csstype.Percentage
+import csstype.Properties
+import kotlinx.js.Record
+
+@JsName("keyframes")
+external fun rawKeyframes(blocks: Record<Percentage, Properties>): AnimationName
```
turansky, line 12: block on new line?

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
```diff
@@ -0,0 +1,21 @@
...
+package emotion.css
+
+import csstype.*
+import kotlinx.js.Record
+import kotlinx.js.jso
+
+@CssDsl
+interface KeyframesBuilder {
+    inline operator fun Percentage.invoke(block: Properties.() -> Unit) {
+        unsafeCast<Record<Percentage, Properties>>()[this] = jso(block)
```
turansky, line 14: block on new line?

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
```diff
@@ -0,0 +1,21 @@
+@file:Suppress(
+    "DECLARATION_CANT_BE_INLINED",
```
turansky, line 2: Is it legal?

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
turansky, line 12: Alias required for record

turansky (changes requested):

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
turansky, line 12: Could we use original name?

turansky (changes requested):

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
sgrishchenko: fixed

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
sgrishchenko: fixed

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
sgrishchenko: It was made by analogy with `RuleBuilder`

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
sgrishchenko: do you propose overload?

sgrishchenko: What name do you propose?

sgrishchenko: I think original signature would be very unuseful for end user, DSL overload is more similar to original DSL from emotion

turansky: `Keyframes`?

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
turansky: Cancelled

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
turansky: > do you propose overload?
Yes

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
```diff
@@ -0,0 +1,14 @@
...
+package emotion.css
+
+import csstype.AnimationName
+import csstype.Percentage
+import csstype.Properties
+import kotlinx.js.Record
+
+@JsName("keyframes")
+external fun rawKeyframes(
+    blocks: Record<Percentage, Properties>,
```
turansky, line 13: `block`?
`Record<String, Properties>`?

turansky (changes requested):

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
sgrishchenko: I used `blocks` because in MDN specs this fragment is named `<keyframe-block-list>`, do you think just `block` would be better?
`Record<String, Properties>` is incorrect alias, according to specs only these values are available `<keyframe-selector> = from | to | <percentage>`, I have supported `from` and `to` additionally

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
sgrishchenko: but what about my point? Do you think it is not a problem?

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
sgrishchenko: `blocks` is renamed

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.kt
sgrishchenko: after discussion it was decided to keep original signature public because it does not conflict with extension

turansky (approved):

PR #1306, opened 2022-03-07
Emotion. Global API is fixed

sgrishchenko:
cc @turansky @aerialist7

### DIFF
--- a/kotlin-emotion/src/main/kotlin/emotion/react/Global.ext.kt
+++ b/kotlin-emotion/src/main/kotlin/emotion/react/Global.ext.kt
@@ -1,10 +1,12 @@
 package emotion.react
 
 import csstype.PropertiesBuilder
+import csstype.RuleBuilder
+import csstype.Rules
 import kotlinx.js.jso
 
 inline fun GlobalProps.styles(
-    crossinline block: PropertiesBuilder.() -> Unit,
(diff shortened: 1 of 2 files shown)


turansky (approved):

aerialist7 (approved):

kotlin-emotion/src/main/kotlin/emotion/react/Global.ext.kt
```diff
@@ -1,10 +1,12 @@
 package emotion.react
 
 import csstype.PropertiesBuilder
+import csstype.RuleBuilder
+import csstype.Rules
 import kotlinx.js.jso
 
 inline fun GlobalProps.styles(
-    crossinline block: PropertiesBuilder.() -> Unit,
+    crossinline block: RuleBuilder<PropertiesBuilder>.() -> Unit,
```
turansky, line 9: What about pseudos?

turansky (changes requested):

turansky (approved):

kotlin-emotion/src/main/kotlin/emotion/react/Global.ext.kt
sgrishchenko: pseudos should be linked with some selector, they can be used on top level

PR #1312, opened 2022-03-08
Emotion. Fix DCE in `KeyframeBuilder`

stefanthaler:
cc @turansky

### DIFF
--- a/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
+++ b/kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
@@ -10,28 +10,39 @@ import kotlinx.js.jso
 
 typealias Keyframes = Record<Percentage, Properties>
 
+private val FROM_PERCENTAGE = 0.pct
+private val TO_PERCENTAGE = 100.pct
+
 @CssDsl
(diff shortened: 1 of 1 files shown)


kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
```diff
@@ -10,28 +10,39 @@ import kotlinx.js.jso
...
+    fun build(): Keyframes = keyframes
 }
 
 inline fun keyframes(
     block: KeyframesBuilder.() -> Unit,
 ): AnimationName =
-    keyframes(jso(block).unsafeCast<Keyframes>())
+    keyframes(
+        KeyframesBuilder()
+            .apply { block() }
```
turansky, line 46: `.apply(block)`?

turansky (changes requested):

kotlin-emotion/src/main/kotlin/emotion/css/keyframes.ext.kt
stefanthaler: Done

turansky (approved):

PR #1315, opened 2022-03-08
Emotion. Styled Components can be used as selectors

sgrishchenko:
cc @turansky @aerialist7 @stefanthaler


kotlin-emotion/src/main/kotlin/emotion/styled/styled.ext.kt
```diff
@@ -14,12 +14,18 @@ private inline val Props.theme: Theme
 fun <P : PropsWithClassName> ElementType<P>.styled(
     options: (StyledOptions.() -> Unit)? = null,
```
turansky, line 15: Redundant?

kotlin-emotion/src/main/kotlin/emotion/styled/generateId.kt
```diff
@@ -0,0 +1,15 @@
+package emotion.styled
+
+import emotion.hash.hash
+import react.ElementType
+
+private val identifiers = mutableMapOf<String, Int?>()
+
+internal fun generateId(component: ElementType<*>): String {
+    val displayName: String? = component.asDynamic().displayName
+    val name = displayName?.let { escape(it) } ?: "Anonymous"
```
turansky, line 10: Anonymous for intrinsic?

kotlin-emotion/src/main/kotlin/emotion/hash/hash.kt
```diff
@@ -0,0 +1,7 @@
+@file:JsModule("@emotion/hash")
```
turansky, line 1: Do we need one more dependency?

turansky (changes requested):

kotlin-emotion/src/main/kotlin/emotion/styled/styled.ext.kt
sgrishchenko: `shouldForwardProp` can be useful

kotlin-emotion/src/main/kotlin/emotion/styled/generateId.kt
sgrishchenko: It is removed

turansky (approved):

PR #1359, opened 2022-03-24
CSSType. `GridArea` factory method

stefanthaler:
See: https://developer.mozilla.org/en-US/docs/Web/CSS/grid-area

cc @turansky


turansky (changes requested):
Generated factories required instead with different combinations

PR #1368, opened 2022-03-27
MUI: Correct Palette and PaletteOptions types

xian:
Added types for most of the remaining `Palette` and `PaletteOptions` fields.

@turansky — I realize these files are generated, not hand-edited, so this PR won't be directly acceptable, but I figured this was the easiest way to convey suggested changes to [mui-kotlin](https://github.com/karakum-team/mui-kotlin).

Thank you!


turansky:
@xian You can mark it draft to avoid accept

> but I figured this was the easiest way to convey suggested changes

Yes, it will be helpful.
Thank you!

turansky (changes requested):
Mark it as draft please

PR #1386, opened 2022-04-02
JS. `HighResTimeStamp`

turansky:
[Documentation](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp)

cc @sgrishchenko @stefanthaler @aerialist7


aerialist7 (commented):
Why not `DOMHighResTimeStamp`?

turansky:
Because it's not browser specific

aerialist7 (approved):

PR #1410, opened 2022-04-07
React-Query. Add safe `QueryKey` factories

stefanthaler:
cc: @turansky

### DIFF
--- a/kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
+++ b/kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
@@ -1,0 +1,13 @@
+@file:Suppress(
+    "NOTHING_TO_INLINE"
+)
+
+import react.query.QueryKey
+
+inline fun <T : QueryKey> QueryKey(
+    key: String,
+): QueryKey = key
+
+inline fun <T : QueryKey> QueryKey(
+    vararg keys: Any,
+): QueryKey = keys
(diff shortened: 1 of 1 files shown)


kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
```diff
@@ -0,0 +1,13 @@
+@file:Suppress(
+    "NOTHING_TO_INLINE"
```
turansky, line 2: Missed trailing comma

kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
```diff
@@ -0,0 +1,13 @@
...
+)
+
+import react.query.QueryKey
```
turansky, line 5: Redundant import?

turansky (changes requested):

kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
stefanthaler: Fixed.

kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
stefanthaler: No, seems to be necessary.

turansky: Because `package` missed? :)

turansky (changes requested):

kotlin-react-query/src/main/kotlin/react/query/QueryKey.ext.kt
stefanthaler: Fixed.

turansky (approved):

PR #1427, opened 2022-04-14
Browser. Fix event type [generated]

turansky:
cc @stefanthaler @aerialist7


aerialist7 (approved):
No more double clicked ☝️

PR #1486, opened 2022-05-11
Fix BOM usage sample

aerialist7:


turansky (changes requested):
BOM readme update required too

turansky (approved):

PR #1492, opened 2022-05-15
[kotlin-styled-next] Dynamic CSS feature

### LINKED ISSUE #1477: Dynamic styles with static human-readable class names [kotlin-styled-next]
This can be a logical continuation of #356. There is a support for static style sheets provided by the `StyleSheet` class. It's also possible to declare dynamic CSS rules right inside components: these rules can have conditional branches and provide different styling depending on some context. But these dynamic rules will get auto-generated names (i.e. `ksc-<n>`) which are not very clear to investigate / debug your styling.

It would be nice to have a possibility to name even such dynamic style classes. These names should be generated according to the context (arguments) required to infer a concrete style, i.e., `(argument: T) -> RuleSet`, a particular `RuleSet` is named as `<StyleSheetClass>-<basicStyleName>-<contextName>`.

1. Manual CSS manipulation without any additional libraries: will not do as it requires lots of work which is already done.
2. Creation of multiple static `StyleSheet`s and branching between their styles inside components' code: can do, but is not very flexible and requires bigger number of style classes created.
3. Extend the existing `StyleSheet` to provide the described solution. See **Additional context** for possible example.

**Additional context**
Here is my current solution (see the attached archive with sources). It was developed and works with **kotlin-styled-next:1.1.0**. An example of usage:
```kotlin
private object NotificationHeaderStyles : DynamicStyleSheet() {

    val container: DynamicCssProvider<Context> by dynamicCss {
        display = Display.flex
        backgroundColor = Theme.palette.backSpecial1(it)
    }

    val titleAndDescriptionRows: RuleSet by css { flexGrow = 1.0 }
    
}
```
and inside of some component:
```kotlin
val context = useAppContext()
styledDiv {
    css(NotificationHeaderStyles.container(context))
}
```
Each object used as an argument for dynamic style must extend:
```kotlin
interface HasCssSuffix {
    val cssSuffix: String
}
```
to properly name each dynamic style. Note that my proposal implementation is a kind of raw initial version: there were some `internal` methods in the original `StyleSheet` which are not very clear to me (as could not find any of their usages) so they were dropped.
[dynamic-css-proposal.tar.gz](https://github.com/JetBrains/kotlin-wrappers/files/8646564/dynamic-css-proposal.tar.gz)

andrew-k-21-12:
This PR can be treated as an initial attempt to demonstrate the possible solution and close #1477 - all descriptions are inside this issue.

The demo project - kotlin-styled-next-dynamic-css-demo - can be obviously dropped from the sources (just needed some way to test updates).

### DIFF
--- a/kotlin-styled-next/src/main/kotlin/styled/HasCssSuffix.kt
+++ b/kotlin-styled-next/src/main/kotlin/styled/HasCssSuffix.kt
@@ -1,0 +1,13 @@
+package styled
+
+/**
+ * Provides some [cssSuffix] to be used in construction of CSS class names.
+ * Simple hash codes are not suitable for this as they are not very human-readable
+ * and their computation can be more complex than needed.
+ * */
+interface HasCssSuffix {
+
+    /** A suffix to be used in name constructions of CSS classes. */
+    val cssSuffix: String
+
+}
--- a/kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
+++ b/kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
@@ -36,11 +36,23 @@ open class StyleSheet(
         }
     private val holders: MutableList<CssHolder> = mutableListOf()
 
+    /** Keeps all holders providing dynamic styles - the holders are cached by their CSS suffixes. */
+    private val dynamicHolders = mutableMapOf<String, DynamicCssHolder>()
+
     constructor(name: String, parent: StyleSheet, isStatic: Boolean = true) : this(parent.name + "-" + name, isStatic)
 
     fun css(vararg parents: RuleSet, builder: RuleSet) =
         CssHolder(this, *parents, builder)
             .also { addCssHolder(it) }
+
+    /**
+     * Declares properties with dynamic styles according to provided arguments.
+     *
+     * @param builder Describes how to prepare a style according to the current argument.
+     *
+     * @return An instance of the [DynamicCssDelegate] allowing the target property to invoke required styles.
+     * */
+    fun <T : HasCssSuffix> dynamicCss(builder: CssBuilder.(T) -> Unit) = DynamicCssDelegate(this, builder)
 
     internal fun addCssHolder(holder: CssHolder) {
         holders.add(holder)
@@ -49,6 +61,28 @@ open class StyleSheet(
     fun inject() {
         scheduleToInject()
         GlobalStyles.injectScheduled()
+    }
+
+    /**
+     * Creates a new or uses an already cached [RuleSet] corresponding to the provided [argument].
+     *
+     * @param staticCssSuffix Some unique static (keeping even if the [argument] changes) name for a style:
+     *                        usually a property name.
+     * @param builder Describes how to prepare styles for the particular [argument].
+     * @param argument Some kind of seed and identifier to prepare a [RuleSet].
+     *
+     * @return A prepared [RuleSet] ready to be used.
+     * */
+    internal fun <T : HasCssSuffix> prepareCachedRuleSet(
+        staticCssSuffix: String,
+        builder: CssBuilder.(T) -> Unit,
+        argument: T
+    ): RuleSet {
+        val fullCssSuffix = "$staticCssSuffix-${argument.cssSuffix}"
+        return dynamicHolders.getOrPut(fullCssSuffix) {
+            DynamicCssHolder(this, fullCssSuffix, { builder.invoke(this, argument) })
+                .also { it.markToInject() }
+        }.provideRuleSet()
     }
 
     internal fun scheduleImports() {
--- a/kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
+++ b/kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
@@ -97,16 +97,23 @@ open class StyleSheet(
         holders.forEach {
             it.scheduleToInject()
         }
+        dynamicHolders.values.forEach {
+            it.scheduleToInject()
+        }
     }
 
     fun scheduleToInject(className: String) {
         holders.forEach {
             it.scheduleToInject(className)
         }
+        dynamicHolders.values.forEach {
+            it.scheduleToInject(className)
+        }
     }
 
     fun removeInjected() {
         holders.forEach { it.removeInjected() }
+        dynamicHolders.values.forEach { it.removeInjected() }
     }
 }
 
--- a/kotlin-styled-next-dynamic-css-demo/build.gradle.kts
+++ b/kotlin-styled-next-dynamic-css-demo/build.gradle.kts
@@ -1,0 +1,15 @@
+plugins {
+    kotlin("js")
+}
+
+kotlin {
+    js(IR) {
+        binaries.executable()
+        browser {}
+    }
+}
+
+dependencies {
+    implementation(project(":kotlin-react"))
+    implementation(project(":kotlin-styled-next"))
+}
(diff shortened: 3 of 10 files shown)


Leonya:
I like it but it seems that requiring `HasCssSuffix` is a bit too limiting. We could simply do `if (x is Enum<*>) x.name` instead. Other types beside enum classes could also be supported in this fashion.

andrew-k-21-12:
Got it, thank you. Will do an update. I think it should have both enum- and other types-based protocols. The reason for `HasCssSuffix` is that sometimes (at least in my use cases) some classes more complex than enums can provide a context to retrieve a particular style, i.e.:
```kotlin
data class Context(val screenSize: ScreenSize, val colorMode: ColorMode) : HasCssSuffix {

    // ...
    
    override val cssSuffix: String = "${screenSize.name.lowercase()}${colorMode.name.camelCaseWord()}"

}
```

Leonya:
Sure, just don't overcomplicate it. Let `argument` be of type `Any` and then just handle various types using `when`.

```when (argument)`

**Comment by Leonya** [comment]
https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape

File: kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt [hunk 1]
```diff
@@ -73,18 +74,29 @@ open class StyleSheet(
...
+    private fun getArgumentCssSuffix(argument: Any): String = when (argument) {
+        is Boolean -> argument.toString()
+        is Number -> argument.toString().replace(".", "-")
+        is String -> argument.revampCssSuffix()
+        is Enum<*> -> argument.name.revampCssSuffix()
+        is HasCssSuffix -> argument.cssSuffix.revampCssSuffix()
+        else -> throw IllegalArgumentException("type is unsupported")
+    }
+
+    private fun String.revampCssSuffix() = CSS.escape(this.replace(" ", ""))
```
Leonya, line 98: `this.replace(" ", "")` is not necessary, escape does all the necessary escaping.

`revamp` -> `escape`

andrew-k-21-12:
Have done the updates mostly in the way you proposed. A couple of differences:
- numbers can have dots which harm loading of CSS classes in HTML, `CSS.escape(...)` doesn't help here
- `CSS.escape(...)` didn't help me with spaces in strings, so had to add the corresponding replacement
- there is a super small chance that someone can shoot his feet by providing spaces in enum names (by using back quotes) or as `HasCssSuffix::cssSuffix` - so added revamping here as well

kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
```diff
@@ -73,18 +74,29 @@ open class StyleSheet(
...
                 .also { it.markToInject() }
         }.provideRuleSet()
     }
 
+    private fun getArgumentCssSuffix(argument: Any): String = when (argument) {
+        is Boolean -> argument.toString()
+        is Number -> argument.toString().replace(".", "-")
+        is String -> argument.revampCssSuffix()
+        is Enum<*> -> argument.name.revampCssSuffix()
+        is HasCssSuffix -> argument.cssSuffix.revampCssSuffix()
```
Leonya, line 94: `HasCssSuffix` should go before `Enum<*>`

kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
```diff
@@ -73,18 +74,29 @@ open class StyleSheet(
...
+        val fullCssSuffix = "$staticCssSuffix-${getArgumentCssSuffix(argument)}"
         return dynamicHolders.getOrPut(fullCssSuffix) {
             DynamicCssHolder(this, fullCssSuffix, { builder.invoke(this, argument) })
                 .also { it.markToInject() }
         }.provideRuleSet()
     }
 
+    private fun getArgumentCssSuffix(argument: Any): String = when (argument) {
+        is Boolean -> argument.toString()
+        is Number -> argument.toString().replace(".", "-")
```
Leonya, line 91: `replace(".", "-")` -> `CSS.escape`

Leonya (changes requested):

Leonya:
> numbers can have dots which harm loading of CSS classes in HTML, CSS.escape(...) doesn't help here

Why not?

> CSS.escape(...) didn't help me with spaces in strings, so had to add the corresponding replacement

Why not? Escaping spaces as `\\ ` seems to be legit.

kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
andrew-k-21-12: Yes, sure, fixed.

andrew-k-21-12:
Regarding dots in numbers. Try to construct the following dynamic style:
```kotlin
val numberEscaping by dynamicCss<Number> {
    fontSize = 1.5.rem
}
```
and use it like this:
```kotlin
css(DemoComponentStyles.numberEscaping(1.234))
```
With only `CSS.escape(...)` it ends with something like this - no styles are loaded in a browser:
no-replace
But with `replace(".", "-")` everything is fine.

The same with spaces. If we remove `this.replace(" ", "")` in `String.revampCssSuffix()` and use `CSS.escape(...)` only, styles are not loaded in a browser:
strings
when using the following example CSS:
```kotlin
css(DemoComponentStyles.numberEscaping("got spaces"))
```

Maybe I'm doing something wrong.

Leonya:
Oh, I see. Alright then.

Leonya (approved):

tretikoff:
Hi! Sorry for the long wait. @andrew-k-21-12 I've refactor a bit, please check

tretikoff:
Overall idea is good, but I think that creating the whole new project in wrappers to demo dynamic-css might be an overkill. @Leonya what do you think? Maybe just add tests in `kotlin-styled-next` package?

Leonya:
Yes, of course, the demo project should be refactored to tests.

tretikoff (approved):

andrew-k-21-12 (commented):
@tretikoff Thank you for the refactoring, it was very nice to extract duplicated methods into the `AbstractCssHolder`, just added a couple of super minor suggestions.

andrew-k-21-12:
It seems like it was accidentally merged, as the demo sample is not removed.

Leonya:
Not accidentally, I will remove it in a follow-up commit.

Leonya:
Released in `pre.339`

kotlin-styled-next/src/main/kotlin/styled/StyleSheet.kt
```diff
@@ -60,63 +96,19 @@ open class StyleSheet(
 
     private fun scheduleToInject() {
         scheduleImports()
-        holders.forEach {
+        (holders + dynamicHolders.values).forEach {
```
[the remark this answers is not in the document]
tretikoff: Thought about it, but it's too much holder list properties. Could be hard to navigate

kotlin-styled-next/src/main/kotlin/styled/HasCssSuffix.kt
```diff
@@ -6,8 +6,6 @@ package styled
  * and their computation can be more complex than needed.
  * */
 interface HasCssSuffix {
```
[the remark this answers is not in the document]
andrew-k-21-12: A small note on the package name - it seems like `dynamicCss` [is fine](https://kotlinlang.org/docs/coding-conventions.html#naming-rules), just used to [another style guide](https://developer.android.com/kotlin/style-guide#package_names) 😅

PR #1502, opened 2022-06-01
Coroutines `1.6.2`

turansky:


Leonya:
It doesn't work, I tried it already. There will be a new version soon.

Issue #1403, opened 2022-04-06
Support for MUI Data Grid


dioxic:
Any plans to wrap MUI Data Grid in the future?

https://mui.com/components/data-grid/

turansky:
It looks like next issue after MUI `5.6.0` and "X Pickers" :)

dioxic:
Awesome, thank you sir 😄

turansky:
Open issue will be fine :)

turansky:
@dioxic Do you have list of required MUI Data Grid features? Or list of features, which you use?

dioxic:
Most desired features from MIT version:

- column editing
- filtering
- sorting
- pagination
- single row / checkbox selection

Virtualization is the most desired feature from the commercial version.

PR #1516, opened 2022-06-07
Remove MUIX Data Grid subproject

aerialist7:
cc @turansky 
related #1403


alexstaeding:
I don't know if I'm missing something obvious, but I don't see why this was removed. #1403 apparently tracks this feature, then #1414 merges the feature but does not close the issue. Later, #1516 (this PR) removes it entirely and does not update the original issue.

I couldn't find any explanation anywhere.

turansky:
> then #1414 merges the feature but does not close

It was only subproject.
Feature isn't ready - that is why subproject was removed.

PR #1517, opened 2022-06-07
React Router. Fix imports and nullability

turansky:
cc @sgrishchenko @aerialist7

### DIFF
--- a/kotlin-react-router-dom/src/main/generated/react/router/NavigationContext.kt
+++ b/kotlin-react-router-dom/src/main/generated/react/router/NavigationContext.kt
@@ -5,4 +5,5 @@
 
 package react.router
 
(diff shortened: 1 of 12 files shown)


aerialist7 (approved):

kotlin-react-router-dom/src/main/generated/react/router/NavigationContext.kt
```diff
@@ -5,4 +5,5 @@
 
 package react.router
 
+@JsName("UNSAFE_NavigationContext")
 external val NavigationContext: react.Context<NavigationContextObject>
```
sgrishchenko, line 9: it would be great to rename these context, because now it is not obvious that it is delicate API

PR #1528, opened 2022-06-12
React. Add `useRequiredContext` hook

turansky:
Related - #1526

сс @Leonya @aerialist7


sgrishchenko:
Discussion is needed

sgrishchenko:
More strict implementation
```kt
fun <T : Any> useRequiredContext(
    context: Context<T?>,
): T {
    val value = useContext(context)
    return requireNotNull(value)
}
```

image

aerialist7 (approved):

PR #1531, opened 2022-06-13
Remove `kotlin-styled`

turansky:
cc @Leonya @sgrishchenko @aerialist7


aerialist7 (approved):
So now we must use `emotion`, OK

Leonya:
I think it's premature, it works just fine for now. Let's just put a notice that it's a legacy project. We can remove it when and if it will be hard to support future versions of styled-components.

PR #1544, opened 2022-06-22
Add `hydrateRoot` symbol to `react-dom/client` (#1543)

### LINKED ISSUE #1543: `react-dom`: Need `hydrateRoot` symbol
When serving React on the client-side which has been pre-rendered by a server through SSR (via `react.dom.server.render*`), the client is expected to call [`hydrateRoot`](https://reactjs.org/docs/react-dom-client.html#hydrateroot), a new API in React 18.x.

[`createRoot`](https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-react-dom/src/main/kotlin/react/dom/client/createRoot.kt) already exists and works great, but there is no equivalent call for `hydrateRoot`.

For now, applications can still call `react.dom.hydrate(...)`, which is deprecated at yields the console error:
```
Warning: ReactDOM.hydrate is no longer supported in React 18. Use hydrateRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot
```

To access the new functionality, a shim declaration can be placed in the package `react.dom.client`, like so:
```kotlin
@file:JsModule("react-dom/client")
@file:JsNonModule

package react.dom.client

import org.w3c.dom.Element
import react.ReactNode

// See https://reactjs.org/docs/react-dom.html

// 18.0+
external fun hydrateRoot(
  container: Element,
  children: ReactNode?,
  options: RootOptions = definedExternally,
): Root
```

**Additional context**
I plan to submit a PR proposing the above declaration shortly. Also, thank you for this awesome suite of libraries! :)

sgammon:
This change adds the `hydrateRoot` symbol from `react-dom/client`, which CSR React code can use to co-operate with code served by `react-dom/server`.

For info on usage, see:
https://reactjs.org/docs/react-dom-client.html#hydrateroot

Fixes and closes #1543.


kotlin-react-dom/src/main/kotlin/react/dom/client/hydrateRoot.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.w3c.dom.Element
+import react.ReactNode
+
+// See https://reactjs.org/docs/react-dom.html
+
+// 18.0+
+external fun hydrateRoot(
+  container: Element,
+  children: ReactNode?,
```
turansky, line 14: [Original name](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f589c8f76383f3fab94631052080ae2a81c88305/types/react-dom/client.d.ts#L51) is preferred

turansky (changes requested):

kotlin-react-dom/src/main/kotlin/react/dom/client/hydrateRoot.kt
sgammon: @turansky fixed, and added `HydrationOptions` as well to match [this interface](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f589c8f76383f3fab94631052080ae2a81c88305/types/react-dom/client.d.ts#L8-L14), which may change over time.

PR #1550, opened 2022-06-24
JS `import`

turansky:
Unlocked in [Kotlin 1.7.0](https://kotlinlang.org/docs/whatsnew17.html#dynamically-load-javascript-modules-from-js-expressions)

cc @sgrishchenko @aerialist7

### DIFF
--- a/kotlin-js/src/main/kotlin/kotlinx/js/Module.kt
+++ b/kotlin-js/src/main/kotlin/kotlinx/js/Module.kt
@@ -1,0 +1,5 @@
+package kotlinx.js
+
+external interface Module<out T : Any> {
+    val default: T
+}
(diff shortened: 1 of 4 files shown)


kotlin-js/src/main/kotlin/kotlinx/js/Module.kt
```diff
@@ -0,0 +1,5 @@
+package kotlinx.js
+
+external interface Module<out T : Any> {
+    val default: T
```
sgrishchenko, line 4: Technically default export can be null

turansky: I see following use cases:
1. Module with default export -> use `Module`
2. Module without default export -> don't use `Module`

aerialist7 (approved):

PR #1554, opened 2022-06-29
Browser. `sx` - extension for strict style manipulation

turansky:
cc @sgrishchenko @aerialist7

### DIFF
--- a/kotlin-browser/build.gradle.kts
+++ b/kotlin-browser/build.gradle.kts
@@ -6,4 +6,5 @@ plugins {
 
 dependencies {
     api(project(":kotlin-js"))
+    api(project(":kotlin-csstype"))
 }
--- a/kotlin-browser/src/main/kotlin/org/w3c/dom/css/sx.kt
+++ b/kotlin-browser/src/main/kotlin/org/w3c/dom/css/sx.kt
@@ -1,0 +1,13 @@
+package org.w3c.dom.css
+
+import csstype.CssDsl
+import csstype.Properties
+
+inline val ElementCSSInlineStyle.sx: Properties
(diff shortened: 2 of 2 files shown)


kotlin-browser/src/main/kotlin/org/w3c/dom/css/sx.kt
```diff
@@ -0,0 +1,13 @@
+package org.w3c.dom.css
+
+import csstype.CssDsl
+import csstype.Properties
+
+inline val ElementCSSInlineStyle.sx: Properties
```
aerialist7, line 6: Why not `style` instead of `sx`?

sgrishchenko: `sx` is style extension, it means you can use selectors inside, here is just regular style as I can see it

Leonya:
I thought `kotlin-browser` was for browser APIs, `sx` is not one.

turansky:
> I thought `kotlin-browser` was for browser APIs, `sx` is not one.

Moved

kotlin-browser/build.gradle.kts
```diff
@@ -6,4 +6,5 @@ plugins {
 
 dependencies {
     api(project(":kotlin-js"))
+    api(project(":kotlin-csstype"))
```
Leonya, line 9: This dependency is not needed

Leonya (changes requested):

kotlin-browser/build.gradle.kts
turansky: Fixed

Leonya (approved):

PR #1555, opened 2022-06-29
Browser. `TagName`

turansky:
cc @sgrishchenko @aerialist7

### DIFF
--- a/kotlin-browser/src/main/kotlin/org/w3c/dom/HTML.kt
+++ b/kotlin-browser/src/main/kotlin/org/w3c/dom/HTML.kt
@@ -1,0 +1,351 @@
+package org.w3c.dom
+
+object HTML {
+    inline val a: TagName<HTMLAnchorElement>
+        get() = "a".unsafeCast<TagName<HTMLAnchorElement>>()
+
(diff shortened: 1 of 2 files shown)


aerialist7 (approved):

kotlin-browser/src/main/kotlin/org/w3c/dom/HTML.kt
```diff
@@ -0,0 +1,351 @@
...
+        get() = "var".unsafeCast<TagName<HTMLElement>>()
+
+    inline val video: TagName<HTMLVideoElement>
+        get() = "video".unsafeCast<TagName<HTMLVideoElement>>()
+
+    inline val wbr: TagName<HTMLElement>
+        get() = "wbr".unsafeCast<TagName<HTMLElement>>()
+
+    inline val webview: TagName<HTMLElement>
+        get() = "webview".unsafeCast<TagName<HTMLElement>>()
```
sgrishchenko, line 350: Why we use `unsafeCast` if we have inline factory method for creation of such tag names?

turansky: Fixed in #1558

PR #1566, opened 2022-07-04
React. `displayName` - regular extension property

turansky:
For HOCs and plugins

cc @sgrishchenko @aerialist7


turansky:
We will add sugar in setter after checks

PR #1650, opened 2022-08-03
Fixed generation 0.pct generation to 0%

Windrushfarer:
When using `flex` property, right now we can't set default `flex: 1`, which will be `flex-grow: 1`, `flex-shrink: 1` and `flex-basis: 0%`. That's how most browsers transform `flex: 1` property.
In `kotlin-css` any zero value transforms into `"0"`, which in the browser fallbacks to `0px`. So the `flex(1.0, 1.0, 0.pct)` call will be transformed to `flex: 1 1 0px`. Pixels and percentages have different behavior, so I suggest to keep unit if someone puts `.pct` for zero value too.
I don't know if it's important to keep unit for them, so I made a fix only for percentages.

cc @Leonya


Leonya:
I've added the ability to set one-value and two-value flex: https://github.com/JetBrains/kotlin-wrappers/commit/7e28c612499f35965def7dce6b486f10b8581d08

As a temporary workaround you may use `put("flex", "1")`

Issue #1644, opened 2022-08-02
Integrate AbortController with coroutines


vlsi:
**Is your feature request related to a problem? Please describe.**

The current wrapper for `window.fetch` does not handle cancellation. In other words, if the coroutine cancels, then the fetch keeps running.

**Describe the solution you'd like**

See https://developer.chrome.com/blog/abortable-fetch

It would be nice to integrate `AbortSignal` with coroutines.
For instance:

```kotlin
// Deploy AbortController to the context, and somehow fire the signal when job is cancelled
withContext(AbortSignalContext()) {
    window.abortableFetch(...) /* it would automatically inherit abort signal from coroutine context */
}
```

**Describe alternatives you've considered**

Currently, I use `.asDynamic().signal = ...`, however, it is kind of verbose to use:

```kotlin
val abortController = AbortController()
val response = window.fetch(
    url,
    jso {
        this.method = httpMethod
        this.headers = json(
            "Accept" to "application/json",
            "Content-Type" to "application/json"
        )
        this.signal = abortController.signal // <-- HERE
).let { promise ->
    suspendCancellableCoroutine { cont ->
        cont.invokeOnCancellation { abortController.abort() }
        promise.then(
            onFulfilled = { cont.resume(it) },
            onRejected = { cont.resumeWithException(it) }
        )
    }
}
```

turansky:
* Required classes and extensions (for `RequestInit` too) you can find [here](https://github.com/JetBrains/kotlin-wrappers/tree/master/kotlin-browser/src/main/kotlin/org/w3c/fetch)
* To construct `RequestInit` you can use `jso` (as for other "options" and JSO)

vlsi:
I just realized both window.fetch and AsyncController are a part of kotlin-stdlib-js, so the issue does not belong to kotlin-wrappers

PR #1663, opened 2022-08-06
JS. Common `AbortController` and `AbortSignal`

turansky:
Related - #1644

cc @sgrishchenko @vlsi @aerialist7


aerialist7 (approved):

Ayfri:
Just for information, why emulating a constructor ?

turansky:
We can't declare it as class, because it extends `EventTarget` in browser

But common constructor required for common code, for example for `fetch`

PR #1700, opened 2022-08-13
JS. Suspend `fetch`

### LINKED ISSUE #1644: Integrate AbortController with coroutines

turansky:
Fix #1644

### DIFF
--- a/kotlin-js/build.gradle.kts
+++ b/kotlin-js/build.gradle.kts
@@ -5,5 +5,7 @@ plugins {
 }
 
 dependencies {
+    implementation(kotlinxCoroutines("core"))
+
     testImplementation(kotlin("test-js"))
 }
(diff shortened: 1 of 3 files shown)


turansky:

Issue #1681, opened 2022-08-10
Expand nodejs APIs


vlsi:
Please add

* [x] process (e.g. `process.cwd`, `process.os()`, `process.env[...]`, `process.exitCode`) #1681
* [x] os.homedir() #1685
* [x] path.path #1690
* [x] crypto.createHash #1752
* [x] url.parse #1689
* [x] http.createServer #1719
* [x] http.ServerResponse #1719
* [x] http.IncomingMessage #1719
* [x] node.fs.createReadStream #1684
* [x] node.fs.createWriteStream #1684
* [x] node.fs.readFile #1707
* [x] node.fs.writeFile #1707
* [x] Buffer.alloc #1682
* [x] Buffer.concat #1682
* [x] stream.Duplex #1712
* [x] stream.ReadableStream.on(event, callback) #1712
* [x] stream.ReadableStream.pipe #1720
* [x] stream.WritableStream #1712
* [x] stream.finished #1751
* [x] PipeOptions #1733
* [x] `node.fs.ReadStream`  should implement `node.stream.Readable` #1732
* [x] copyFile(src: PathLike, dest: PathLike) (suspend functions with optional last argument) #1750
* [x] writeFile(file: PathLike, data: Any) (suspend functions with optional last argument) #1750
* [x] readFile(path: PathLike) (suspend functions with optional last argument) #1750
* [x] stat(path: PathLike) (suspend functions with optional last argument) #1750
* [x] global #1749

Just in case, the functions above are required for https://github.com/burrunan/gradle-cache-action

turansky:
Accepted

vlsi:
```
external interface ReadableStream : node.events.IEventEmitter {

    fun <T : WritableStream> pipe(
        destination: T,
        options: Any /* { end?: boolean | undefined; } */ = definedExternally,
    ): T
```

I guess the relevant d.ts is https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f19301b477ff45540091eb108560e1ef61facdc1/types/node/stream.d.ts#L24-L31

```javascript
    class internal extends EventEmitter {
        pipe<T extends NodeJS.WritableStream>(
            destination: T,
            options?: {
                end?: boolean | undefined;
            }
        ): T;
    }
```

It would be nice if `PipeOptions` type could be generated for typesafe `end` property access.

PR #1750, opened 2022-08-22
Node. Additional `suspend` adapters [generated]

turansky:
Part of #1681


kotlin-node/src/main/generated/node/fs/stat.kt
```diff
@@ -4,6 +4,13 @@ package node.fs
 
 import kotlinx.coroutines.await
 
+suspend fun stat(
+    path: PathLike,
+): Any /* Stats | BigIntStats */ =
+    statAsync(
+        path = path,
+    ).await()
```
vlsi, line 12: This is known to return Stats, not BigIntStats

turansky:

Issue #1271, opened 2022-02-26
Add missing WebGL typings


Ayfri:
**Is your feature request related to a problem? Please describe.**
For a project, I need multiple classes/interfaces missing from the standard library, I suggest adding them to `kotlin-browser`.

I require all the `WEBGL_compressed_texture_*` typings, there are multiple interfaces with constants in it.
Also, some interfaces, and classes missing from `org.khronos.webgl`.

I thought about adding all the missing interfaces from WebGL, I listed them all so.

**Describe the solution you'd like**
Add typings to these interfaces :
- `EXT_blend_minmax`
- `EXT_disjoint_timer_query`
- `EXT_color_buffer_half_float`
- `EXT_frag_depth`
- `EXT_sRGB`
- `EXT_shader_texture_lod`
- `EXT_texture_filter_anisotropic`

- `OES_element_index_uint`
- `OES_standard_derivatives`
- `OES_texture_float`
- `OES_texture_float_linear`
- `OES_texture_half_float`
- `OES_texture_half_float_linear`
- `OES_vertex_array_object`

- `WEBGL_color_buffer_float`

- `WEBGL_compressed_texture_atc`
- `WEBGL_compressed_texture_astc`
- `WEBGL_compressed_texture_etc1`
- `WEBGL_compressed_texture_pvrtc`
- `WEBGL_compressed_texture_s3tc`
- `WEBGL_compressed_texture_s3tc_srgb`

- `WEBGL_debug_renderer_info`
- `WEBGL_debug_shaders`
- `WEBGL_depth_texture`
- `WEBGL_draw_buffers`
- `WEBGL_lose_context`

Add typings to these classes :
- `WebGL2RenderingContext`
- `WebGLQuery`
- `WebGLSampler`
- `WebGLSync`
- `WebGLTransformFeedback`
- `WebGLVertexArrayObject`

**Describe alternatives you've considered**
Implement myself for my project the typings, which is what I'm doing right now.

turansky:
@Ayfri Could you please mark converted/non-converted types?
Example - #1681

Ayfri:
Sorry, but what do you mean by "converted" ?

turansky:
"converted" = "generated"
For example `WebGLQuery ` already generated

Ayfri:
Done :+1:

turansky:
`EXT_disjoint_timer_query` and `WEBGL_compressed_texture_pvrtc` are unsupported according MDN.
[...]

PR #1771, opened 2022-08-28
WebGL. Classes [generated]

turansky:
Part of #1271


turansky:

PR #1836, opened 2022-10-12
Update declarations [generated]

porotkin:
cc @turansky, @aerialist7


aerialist7 (changes requested):
Why is this done in a single PR?

PR #1841, opened 2022-10-15
Tanstack Query Devtools. Initial

porotkin:


kotlin-tanstack-react-query-devtools/src/main/kotlin/tanstack.react.query.devtools/ReactQueryDevtools.kt
```diff
@@ -0,0 +1,12 @@
+@file:JsModule("@tanstack/react-query-devtools")
```
turansky, line 1: Missed `JsNonModule`?

turansky, line 1: Invalid folder?

turansky (changes requested):

PR #1852, opened 2022-11-16
React Router internal library is added

sgrishchenko:
cc @turansky


turansky (changes requested):
Registration in BOM missed

turansky (changes requested):
Common README also must be updated

PR #1860, opened 2022-11-19
React Router: karakum.config.json is improved

sgrishchenko:


turansky:
Where is valid folders (according packages)? :)

sgrishchenko:
What do you mean saying "valid folders"? Folders for custom plugins? If so it is 
```
buildSrc/karakum/**/*.js
<projectDir>/karakum/**/*.js
```
by default but it is cofigurable

Leonya:
The package is `remix.run.router`, so the path should be `src/main/generated/remix/run/router`.

sgrishchenko:
I can fix it because of conversions in this repo but FYI https://kotlinlang.org/docs/coding-conventions.html#directory-structure
image

sgrishchenko:
I have fixed file structure BTW.

PR #1878, opened 2023-01-15
React Router: router declarations are added

sgrishchenko:


kotlin-remix-run-router/src/main/generated/remix/run/router/RevalidationState.kt
```diff
@@ -0,0 +1,11 @@
...
+@file:JsNonModule
+
+@file:Suppress(
+    "NON_EXTERNAL_DECLARATION_IN_INAPPROPRIATE_FILE",
+)
+
+package remix.run.router
+
+
+typealias RevalidationState = Any /* "idle" | "loading" */
```
turansky, line 11: Conversion for unions isn't implemented?

sgrishchenko: It is marked in TODO list in README for future implementation

Leonya:
This commit contains `Router.kt` and `router.kt` in the same directory, this should be fixed.

PR #1880, opened 2023-01-20
Remix Router v1.3.0

sgrishchenko:


kotlin-remix-run-router/src/main/generated/remix/run/router/DeferredData.kt
```diff
@@ -10,25 +10,30 @@ package remix.run.router
...
     var abortPromise: Any? /* some expression */
     var unlistenAbortSignal: Any? /* some expression */
-    var subscriber: Any? /* some expression */
+    var subscribers: Any? /* some expression */
     var data: Record<String, Any>
+    var init: ResponseInit?
+    var deferredKeys: Array<String>
 
-    constructor (data: Record<String, Any>)
+    constructor (data: Record<String, Any>, responseInit: ResponseInit = definedExternally)
```
turansky, line 27: Why constructor isn't first?

kotlin-remix-run-router/src/main/generated/remix/run/router/DeferredData.kt
```diff
@@ -10,25 +10,30 @@ package remix.run.router
...
 
     var trackPromise: Any? /* some expression */
     var onSettle: Any? /* some expression */
-    fun subscribe(fn: (aborted: Boolean) -> Unit): Unit
+    var emit: Any? /* some expression */
+    fun subscribe(fn: (aborted: Boolean, settledKey: String) -> Unit): () -> Boolean
     fun cancel(): Unit
     fun resolveData(signal: AbortSignal): Promise<Boolean>
     /* get done(): boolean; */
```
turansky, line 35: `val`?

kotlin-remix-run-router/src/main/generated/remix/run/router/createStaticHandler.kt
```diff
@@ -0,0 +1,11 @@
...
+@file:JsNonModule
+
+@file:Suppress(
+    "NON_EXTERNAL_DECLARATION_IN_INAPPROPRIATE_FILE",
+    "NAME_CONTAINS_ILLEGAL_CHARS",
+)
+
+package remix.run.router
+
+external fun createStaticHandler(routes: Array<AgnosticRouteObject>, opts: CreateStaticHandlerOpts = definedExternally): StaticHandler
```
turansky, line 11: `ReadonlyArray`?

kotlin-remix-run-router/src/main/generated/remix/run/router/BlockerUnblocked.kt
```diff
@@ -0,0 +1,18 @@
...
+    "NON_EXTERNAL_DECLARATION_IN_INAPPROPRIATE_FILE",
+    "NAME_CONTAINS_ILLEGAL_CHARS",
+)
+
+package remix.run.router
+
+
+external interface BlockerUnblocked {
+    var state: String
+    var reset: Any? /* some expression */
```
turansky, line 14: "some expression" can be in comment?

turansky (changes requested):

kotlin-remix-run-router/src/main/generated/remix/run/router/DeferredData.kt
sgrishchenko: It is simpler for automatic conversion

kotlin-remix-run-router/src/main/generated/remix/run/router/createStaticHandler.kt
sgrishchenko: There is no signs in source code that this array should be readonly, I will perform revision and add list of places where it would be great to make arrays readonly.

kotlin-remix-run-router/src/main/generated/remix/run/router/DeferredData.kt
sgrishchenko: There is TODO item in README about getters

kotlin-remix-run-router/src/main/generated/remix/run/router/BlockerUnblocked.kt
sgrishchenko: Right now "some expression" is rendered when there is no type at all or type is `null` or `undefined`, I need to rework some logic in converter to separate these two cases and then I can write some more useful comments in such places.

PR #1883, opened 2023-01-22
React Router: all karakum plugins are moved to special subdirectory

sgrishchenko:
+ `StaticHandler` opts are fixed


kotlin-remix-run-router/src/main/generated/remix/run/router/StaticHandler.kt
```diff
@@ -17,7 +17,7 @@ import kotlin.js.Promise
 
 external interface StaticHandler {
     var dataRoutes: Array<AgnosticDataRouteObject>
-    fun query(request: Request, opts: Opts = definedExternally): Promise<Any /* StaticHandlerContext | Response */>
-    fun queryRoute(request: Request, opts: Opts = definedExternally): Promise<Any?>
+    fun query(request: Request, opts: StaticHandlerQueryOpts = definedExternally): Promise<Any /* StaticHandlerContext | Response */>
```
turansky, line 20: nested interface?

turansky (changes requested):

PR #1894, opened 2023-01-28
React Router: hooks and contexts are converted

sgrishchenko:


kotlin-react-router/src/main/generated/react/router/NonIndexRouteObject.kt
```diff
@@ -0,0 +1,26 @@
...
+    var caseSensitive: Any /* AgnosticNonIndexRouteObject["caseSensitive"] */?
+    var path: Any /* AgnosticNonIndexRouteObject["path"] */?
+    var id: Any /* AgnosticNonIndexRouteObject["id"] */?
+    var loader: Any /* AgnosticNonIndexRouteObject["loader"] */?
+    var action: Any /* AgnosticNonIndexRouteObject["action"] */?
+    var hasErrorBoundary: Any /* AgnosticNonIndexRouteObject["hasErrorBoundary"] */?
+    var shouldRevalidate: Any /* AgnosticNonIndexRouteObject["shouldRevalidate"] */?
+    var handle: Any /* AgnosticNonIndexRouteObject["handle"] */?
+    var index: Boolean?
+    var children: Array<RouteObject>?
```
turansky, line 22: `ReadonlyArray`?

kotlin-react-router/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ plugins {
 dependencies {
     api(project(":kotlin-browser"))
     api(project(":kotlin-remix-run-router"))
+    api(project(":kotlin-react"))
```
turansky, line 10: `kotlin-react-core`?

turansky (changes requested):

kotlin-react-router/karakum/plugins/convertReactQualifiedName.js
```diff
@@ -0,0 +1,21 @@
+const ts = require("typescript");
```
turansky, line 1: Why it can't be `import`?

turansky (changes requested):

kotlin-react-router/karakum/plugins/convertReactQualifiedName.js
sgrishchenko: Because right now I compile my source code of Karakum converter as CommonJS, I will try to migrate to Kotlin or TS for plugins in future.

kotlin-react-router/build.gradle.kts
sgrishchenko: `kotlin-react-core` is syntetic module just for `Space` usage, I don't want to be dependent of it, because it can be removed in futere.

turansky: `kotlin-react-core` is what you need :)

sgrishchenko: What do you mean?

kotlin-react-router/src/main/generated/react/router/NonIndexRouteObject.kt
sgrishchenko: I propose to discuss it separately.

kotlin-react-router/build.gradle.kts
sgrishchenko: I propose to discuss it separately.

PR #1898, opened 2023-02-02
React Router: index access type conversion is improved

sgrishchenko:


kotlin-react-router/src/main/generated/react/router/_renderMatches.kt
```diff
@@ -8,9 +8,6 @@
...
 package react.router
 
+import remix.run.router.RouterState
 
-external fun _renderMatches(
-    matches: Array<RouteMatch<*>>?,
-    parentMatches: Array<RouteMatch<*>> = definedExternally,
-    dataRouterState: Any /* RemixRouter["state"] */ = definedExternally
-): react.ReactElement<*>?
+external fun _renderMatches(matches: Array<RouteMatch<*>>?, parentMatches: Array<RouteMatch<*>> = definedExternally, dataRouterState: RouterState = definedExternally): react.ReactElement<*>?
```
turansky, line 13: Is it private?

turansky (approved):

kotlin-react-router/src/main/generated/react/router/_renderMatches.kt
sgrishchenko: Maybe, I don't know, maybe it is exposed just to reuse other `react-router` packages.

PR #1903, opened 2023-02-05
React Router: new DOM subproject is created

sgrishchenko:


kotlin-react-router-dom/karakum.config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "input": [
+    ],
+    "output": "src/main/generated",
+    "libraryName": "react-router",
```
turansky, line 5: Valid?

turansky (changes requested):

kotlin-react-router-dom/karakum.config.json
sgrishchenko: No, fixed

PR #1915, opened 2023-02-17
Do not trigger CI on push in a non-main branch

sgrishchenko:


.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: CI
 
-on: [ push, pull_request ]
+on:
+  push:
+    branches:
+      - main
```
turansky, line 6: `master`?

Issue #1916, opened 2023-02-17
New React Router issues


turansky:
- [x] Invalid `NavigateFunction` declaration
- [x] Invalid `RouteObject` - single type expected in result
- [x] Duplicated `useRoutes` - remove `definedExternally`
- [x] Missed `LocationState` marker interface
- [x] ~`Navigator.listen`, `Navigator.location` - internal?~
- [x] ~Remove redundant `RouteObject` checks~
- [x] `HistoryRouter` undefined (invalid name)
- [x] `NavigationContext` undefined (invalid name)
- [x] ~links don't work (url changed without action)~
- [x] `sealed` interfaces
- [x] remove legacy subprojects
- [x] strict error types
- [x] ~immutable interfaces are mutable~
- [x] ~Invalid `useOutletContext`?~
- [x] `@nativeInvoke` for functions with multiple signatures

PR #1931, opened 2023-02-22
React Router: hot fix for NavigateFunction is removed

sgrishchenko:
partial fix for #1916

### DIFF
--- a/kotlin-react-router/src/jsMain/generated/react/router/NavigateFunction.kt
+++ b/kotlin-react-router/src/jsMain/generated/react/router/NavigateFunction.kt
@@ -1,8 +1,5 @@
-@file:JsModule("react-router")
-
 @file:Suppress(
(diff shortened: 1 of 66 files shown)


kotlin-react-router/src/jsMain/generated/react/router/NavigateFunction.kt
```diff
@@ -1,3 +1,7 @@
+@file:Suppress(
+    "NON_EXTERNAL_DECLARATION_IN_INAPPROPRIATE_FILE",
```
turansky, line 2: Redundant?

turansky (changes requested):

kotlin-react-router/src/jsMain/generated/react/router/NavigateFunction.kt
sgrishchenko: I will try to improve it in future

PR #1943, opened 2023-03-12
React router: v6.9.0 is automated

sgrishchenko:


settings.gradle.kts
```diff
@@ -1,12 +1,20 @@
 rootProject.name = "kotlin-wrappers"
 
 pluginManagement {
+    repositories {
+        mavenLocal()
```
turansky, line 5: Is it for `karakum` dependencies?

turansky (changes requested):

settings.gradle.kts
```diff
@@ -1,12 +1,20 @@
 rootProject.name = "kotlin-wrappers"
 
 pluginManagement {
+    repositories {
```
turansky, line 4: Looks like local development settings

turansky (changes requested):

settings.gradle.kts
sgrishchenko: I think it is ok, because it is just a prioritized list. I don't know why it is not a default behavior though.

sgrishchenko: It was removed after discussion

PR #1983, opened 2023-04-17
MUI `5.10.6` [generated]

aerialist7:


turansky:
onChange handler declarations are invalid

PR #2003, opened 2023-05-03
Add some documentation for the new React DSL

joffrey-bion:
I did what I could based on my (admittedly quite shallow) understanding of the new DSL.
I followed the initial documentation from `kotlin-react-legacy`, and adapted to the new DSL.

Please feel free to correct anything in there!

### DIFF
--- a/docs/guide/react.md
+++ b/docs/guide/react.md
@@ -1,0 +1,380 @@
+# Kotlin React usage
+
+## Defining an entry point
+
+React needs a root node to render to.
+Use the `createRoot(htmlElement)` function to set the given `htmlElement` as a `Root` element,
+and call `Root.render(reactNode)` to inject the corresponding `ReactNode` as the root:
+
+```kotlin
+fun main() {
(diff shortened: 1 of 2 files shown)


turansky:
@joffrey-bion Could you move guide in `/docs/guide` folder and add link in main README?
In `/docs/guide` you can find existed guides

joffrey-bion:
Oh I was not aware of this folder, thanks. I just followed the example of the `kotlin-react-legacy` module.

The existing 2 guides are pretty small, they fit in one file, and they are module-specific. So why not simply write their contents in the corresponding READMEs? There aren't any links from the READMEs to those guides, so they are not very discoverable currently. I think it would help quite a bit to move that there.

turansky:
> So why not simply write their contents in the corresponding READMEs?

Bacause it's usually more than 1 file

> There aren't any links from the READMEs to those guides, so they are not very discoverable currently

You can add it for `kotlin-react`

PR #2036, opened 2023-06-12
Remove import of the whole react package in styled-next

tretikoff:


kotlin-styled-next/src/jsMain/kotlin/styled/ReactImports.kt
```diff
@@ -0,0 +1,7 @@
+@file:JsModule("react")
+@file:JsNonModule
+
+package styled
+
+@JsName("useInsertionEffect")
+external val supportsInsertionEffect: Boolean?
```
turansky, line 7: internal?

PR #2044, opened 2023-06-15
JS. Add `Promise.toResult` extension

aerialist7:


kotlin-js/src/jsMain/kotlin/js/promise/Promise.kt
```diff
@@ -11,3 +12,9 @@ inline fun <S> Promise.Companion.all(
     values: ReadonlyArray<PromiseResult<S>>,
 ): Promise<ReadonlyArray<S>> =
     all(promise = values.unsafeCast<ReadonlyArray<Promise<S>>>())
+
+suspend fun <T> Promise<T>.asResult(): Result<T> =
```
sgrishchenko, line 16: `Result` instance is just created here from received data and it does not proxy something from `Promise`. I propose to use `toResult` name here.

sgrishchenko (changes requested):

turansky (changes requested):
+1 for `toResult`

kotlin-js/src/jsMain/kotlin/js/promise/Promise.kt
aerialist7: Done

PR #2062, opened 2023-06-26
Css properties

tretikoff:


kotlin-css/src/commonMain/kotlin/kotlinx/css/StyledElement.kt
```diff
@@ -283,6 +290,73 @@ private fun getShorthandValue(
...
+        horizontal
+    )
+}
+
+@Suppress("DuplicatedCode")
+var StyledElement.marginProp: Margin
+    set(value) {
+        declarations["margin"] = value.value
+    }
+    @Deprecated("Write-only property", level = DeprecationLevel.HIDDEN)
```
tretikoff, line 322: Will be available when moving to fully, as we will have `margin by CssProperty`

kotlin-styled-next/src/jsMain/kotlin/styled/StyledComponents.kt
```diff
@@ -45,6 +45,10 @@ interface StyledBuilder<P : PropsWithClassName> {
 }
 
 inline fun StyledBuilder<*>.css(handler: RuleSet) = css.handler()
+inline fun StyledBuilder<*>.css(className: String? = null, handler: RuleSet) {
```
Leonya, line 48: Why nullable?

PR #2084, opened 2023-08-07
MUI. Material `5.13.7` [generated]

aerialist7:


kotlin-mui/src/jsMain/generated/muix/pickers/CalendarPicker.kt
```diff
@@ -6,8 +6,8 @@ package muix.pickers
 
 import js.core.ReadonlyArray
 import js.core.Void
-import js.promise.Promise
```
turansky, line 9: Import was fine

turansky (changes requested):

kotlin-mui/src/jsMain/generated/muix/pickers/CalendarPicker.kt
aerialist7: Done

PR #2092, opened 2023-08-12
Use Gradle's platform instead of enforcedPlatform for BOM

lppedd:
This avoids warnings when publishing a KMP artifact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Just declare `kotlin-wrappers-bom` and specify the modules you need:
 val kotlinWrappersVersion = "1.0.0-pre.612"
 
 dependencies {
-    implementation(enforcedPlatform("org.jetbrains.kotlin-wrappers:kotlin-wrappers-bom:$kotlinWrappersVersion"))
+    implementation(platform("org.jetbrains.kotlin-wrappers:kotlin-wrappers-bom:$kotlinWrappersVersion"))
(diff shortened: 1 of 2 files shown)


turansky (changes requested):
Will it works fine for application?

lppedd:
@turansky I've just tried setting up an executable browser app (with React, using the IJ wizard), and the BOM works as expected with `platform`.

However (I think it's unrelated) beware something was changed in the React types, and I get this:

image

turansky:
Modern browser is already here. :)
[Example of usage](https://github.com/karakum-team/kotlin-react-table-sample/blob/68ff06607e073c0d1132541007b929e9ae1b075f/src/main/kotlin/team/karakum/App.kt#L16).

lppedd:
Replace the imports as per the example and now it works fine. Guess the IJ wizard needs an update.

turansky:
For me wizard is fine in latest IDEA/Kotlin plugin

sgrishchenko:
There is also `kotlin-wrappers-bom/README.md` where `enforcedPlatform`  is mentioned.

sgrishchenko (approved):

sgrishchenko:
We have internal discussion about this change. I will mark this change as draft for a while.

lppedd:
Thanks for the update @sgrishchenko, sure, no problem.

PR #2116, opened 2023-09-15
mui: update Portal.kt JsName

xian:
No symbol at `default`, should be `Portal`.


turansky:
It's generated code, in first line you can find:
```
// Automatically generated - do not modify!
```

Please create an issue instead

cc @aerialist7

PR #2129, opened 2023-09-25
Allow converting Uint8Array-s to Kotlin's types

lppedd:
Not sure if `as-` is the correct way to express what's happening here.  
A new instance is created, so maybe `to-` is better. But then it's not aligned with `Int8Array`.

Edit: on a second thought, while a new instance is created, no copying is performed, the new `Int8Array` is only a view over the buffer, so maybe `as-` is totally ok.

### DIFF
--- a/kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
+++ b/kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
@@ -1,5 +1,9 @@
 @file:Suppress(
     "EXTERNAL_CLASS_CONSTRUCTOR_PROPERTY_PARAMETER",
+    "INLINE_EXTERNAL_DECLARATION",
+    "INLINE_CLASS_IN_EXTERNAL_DECLARATION_WARNING",
+    "WRONG_BODY_OF_EXTERNAL_DECLARATION",
+    "NOTHING_TO_INLINE",
 )
 
 package js.typedarrays
@@ -17,6 +21,12 @@ open external class Uint8Array(
     constructor(length: Int)
     constructor(elements: JsIterable<Byte>)
     constructor(elements: ReadonlyArray<Byte>)
+
+    inline fun toByteArray(): ByteArray =
+        Int8Array(buffer, byteOffset, length).asByteArray()
+
+    inline fun toUByteArray(): UByteArray =
+        toByteArray().asUByteArray()
 
     companion object : TypedArrayCompanion<Uint8Array, Byte>
 }
--- a/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
@@ -13,6 +13,7 @@ tasks.withType<KotlinCompile<*>>().configureEach {
 
         freeCompilerArgs += listOf(
             "-opt-in=kotlin.ExperimentalStdlibApi",
+            "-opt-in=kotlin.ExperimentalUnsignedTypes",
             "-opt-in=kotlin.contracts.ExperimentalContracts",
         )
     }


kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
```diff
@@ -18,5 +22,12 @@ open external class Uint8Array(
     constructor(elements: JsIterable<Byte>)
     constructor(elements: ReadonlyArray<Byte>)
 
+    inline fun asByteArray(): ByteArray =
+        Int8Array(buffer, byteOffset, length).unsafeCast<ByteArray>()
```
turansky, line 26: `unsafeCast` used without reason

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
```diff
@@ -18,5 +22,12 @@ open external class Uint8Array(
     constructor(elements: JsIterable<Byte>)
     constructor(elements: ReadonlyArray<Byte>)
 
+    inline fun asByteArray(): ByteArray =
```
turansky, line 25: `to`?

turansky (changes requested):

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
lppedd: @turansky I need the unsafe cast here, otherwise there is type mismatch.

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
lppedd: @turansky yeah that was my original idea. However, maybe better to keep it the same as for `Int8Array`?

turansky: There is no conversion in `Int8Array.asByteArray`

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
turansky: `Int8Array.asByteArray()`?

lppedd: Addressed.

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
lppedd: Renamed to `toByteArray`.

lppedd:
I'll squash so you can merge in case.

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
```diff
@@ -18,5 +22,12 @@ open external class Uint8Array(
     constructor(elements: JsIterable<Byte>)
     constructor(elements: ReadonlyArray<Byte>)
 
+    inline fun toByteArray(): ByteArray =
+        Int8Array(buffer, byteOffset, length).asByteArray()
+
+    @OptIn(ExperimentalUnsignedTypes::class)
```
turansky, line 28: Looks like candidate for common compiler flag

turansky (changes requested):

kotlin-js/src/jsMain/kotlin/js/typedarrays/Uint8Array.kt
lppedd: @turansky done.

PR #2149, opened 2023-10-08
MUI Base. Remove redundant `@JsName` [generated] 

### LINKED ISSUE #2123: ClickAwayListener woes
I am having difficulties with when I use the `ClickAwayListener` component with recent version of Kotlin Wrappers. It used to work in the past (not sure since when it stopped working), but now I get 

`Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. `

I use it in conjonction with `Popper`, which is getting deprecated. Can this be the cause of the problem? If yes, then the `Popup` (https://mui.com/base-ui/react-popup/) replacement is not yet available. Otherwise, any advise about the direction in which I should hunt?

`pre.618` was last working version. Problem occurs with `pre.619-625`.

aerialist7:
Closes #2123

### DIFF
--- a/kotlin-mui-base/src/jsMain/generated/mui/base/Badge.kt
+++ b/kotlin-mui-base/src/jsMain/generated/mui/base/Badge.kt
@@ -18,5 +18,4 @@ package mui.base
  *
  * - [Badge API](https://mui.com/base-ui/react-badge/components-api/#badge)
  */
-@JsName("Badge")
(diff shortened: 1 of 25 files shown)


turansky (changes requested):
JsName is redundant in these cases

aerialist7:
@Leonya, can you make a release, please?

PR #2151, opened 2023-10-08
MUI X Tree View. Fix named export [generated]

aerialist7:


turansky (changes requested):
JsName is redundant in these cases

PR #2158, opened 2023-10-08
MUI Base `5.0.0-beta.18` [generated]

aerialist7:


kotlin-mui-base/src/jsMain/generated/mui/base/useSlider.types.kt
```diff
@@ -156,24 +156,24 @@ external interface UseSliderReturnValue {
 
     /**
      * Resolver for the hidden input slot's props.
-     * @param otherHandlers props for the hidden input slot
+     * @param externalProps props for the hidden input slot
      * @returns props that should be spread on the hidden input slot
      */
-    var getHiddenInputProps: react.Props /* <TOther extends EventHandlers = {}>(otherHandlers?: TOther) => UseSliderHiddenInputProps<TOther> */
+    var getHiddenInputProps: react.Props /* <ExternalProps extends Record<string, unknown> = {}>(externalProps?: ExternalProps) => UseSliderHiddenInputProps<ExternalProps> */
```
turansky, line 162: Strict handler expected in such cases

turansky (changes requested):

kotlin-mui-base/src/jsMain/generated/mui/base/useSlider.types.kt
aerialist7: What do you mean? 
Here is only change in comment

turansky: Strict type expected instead of comment

aerialist7: Ok, will do it later in separate task. It is only version update with no changes in this place

PR #2164, opened 2023-10-17
Allow putting nullable values in records

sgrishchenko:
I want to align `Record` and `ReadonlyRecord` interfaces with `MutableMap` and `Map` interfaces in Kotlin, and with `JsMap`
`ReadonlyMap`.


turansky:
In case of `Record` aligning = loose strictness.
Additional operators can be used instead.

PR #2168, opened 2023-11-01
CSSType. Container query extensions

aerialist7:


turansky (changes requested):
Looks like it must be in generated block and member (not extension)

PR #2186, opened 2023-12-12
Tanstack React Table. `OnChangeFn`. Support `asFunction` extension

aerialist7:
For the next use case

```kotlin
    jso<TableOptions<Any>> {
        // ...
        onGroupingChange = setGroupingState.asFunction()
        onColumnVisibilityChange = setVisibilityState.asFunction()
    }
```


turansky (changes requested):
1. Why it's extension?
2. Second adapter missed

turansky:
Base problem must be solved instead

PR #2194, opened 2024-01-02
JS. `JsFunction`

sgrishchenko:


turansky (approved):

lppedd:
What's the purpose of introducing `JsFunction`? Wouldn't the same result be accomplished with a Kotlin function/lambda?

sgrishchenko:
It is a special interface to describe such types like `GeneratorFunction` and `AsyncGeneratorFunction`. I would like to use existing `Function` interface, but unfortunately I cannot extend external entities from `Function`.

PR #2204, opened 2024-01-19
React Select. Support `CreatableSelect` component

aerialist7:


kotlin-react-select/src/jsMain/kotlin/react/select/GetOptionLabel2.kt
```diff
@@ -0,0 +1,8 @@
+package react.select
+
+import react.ReactNode
+
+typealias GetNewOptionData<Option> = (
```
turansky, line 5: Invalid file name

kotlin-react-select/src/jsMain/kotlin/react/select/CreatableSelect.ext.kt
```diff
@@ -0,0 +1,11 @@
+package react.select
+
+import js.objects.jso
+import react.ChildrenBuilder
+import react.ReactDsl
+
+fun <Option : Any, Group : GroupBase<Option>> ChildrenBuilder.CreatableSelect(
```
turansky, line 7: Adapter is redundant with new builder

turansky (changes requested):

kotlin-react-select/src/jsMain/kotlin/react/select/GetOptionLabel2.kt
aerialist7: Fixed

kotlin-react-select/src/jsMain/kotlin/react/select/CreatableSelect.ext.kt
aerialist7: I tried to pass generic type directly to `external val CreatableSelect` but it doesn't work.
How it should looks like?

turansky: You can find example in `mui-kotlin` playground

aerialist7: Done, thanks!

turansky (approved):

aerialist7:
@Leonya, could you please make a release?

PR #2239, opened 2024-05-02
MUI. Material `5.15.16` [generated]

aerialist7:


kotlin-mui-material/src/jsMain/generated/mui/material/Select.kt
```diff
@@ -175,7 +175,7 @@ external interface SelectProps<Value> :
      * The variant to use.
      * @default 'outlined'
      */
-    var variant: SelectVariant?
+    var variant: SelectVariants?
```
turansky, line 178: `SelectVariant`? Plural form isn't expected

turansky (changes requested):

kotlin-mui-material/src/jsMain/generated/mui/material/Select.kt
turansky: Single form expected for variant interface name (like in other components)

aerialist7: Why? It's not expected.
In original MUI declarations it calls `SelectVariants`
Here is the usage https://github.com/mui/material-ui/blob/45cf002ee21d0b99f8ba2dc54e4d233499fb689b/packages/mui-material/src/Select/Select.d.ts#L151

turansky (approved):

PR #2240, opened 2024-05-03
MUI. Full style customization support via components' classes [generated]

aerialist7:


turansky:
Could you please attach documentation/example links, which describes update?

turansky (changes requested):
Could you please attach documentation/example links, which describe update?

aerialist7:
This is a fix for correct styling of the components.
See the "Styled components" section here – https://mui.com/base-ui/getting-started/customization/#which-option-to-choose
Here is the sample – https://codesandbox.io/p/sandbox/interesting-oskar-w89gtm?file=%2Fsrc%2FDemo.tsx

There was a problem in MUI Base only, because it has different class-name convention – instead of "MuiComponent-modifier" it uses "base-Component-modifier".

Screenshot 2024-05-03 at 14 23 30

turansky:
`switchClasses` exists in runtime as I see.
External object `switchClasses` expected in declarations.

turansky (changes requested):
`switchClasses` exists in runtime as I see.
External object `switchClasses` expected in declarations.

aerialist7:
> `switchClasses` exists in runtime as I see. External object `switchClasses` expected in declarations.

Done

kotlin-mui-base/src/jsMain/generated/mui/base/Button.classes.kt
```diff
@@ -17,3 +19,10 @@ external interface ButtonClasses {
     /** State class applied to the root `button` element if `focusVisible={true}`. */
     var focusVisible: ClassName
 }
+
+external object buttonClasses : ButtonClasses {
```
turansky, line 23: `external val buttonClasses : ButtonClasses` will work without redundant overrides

turansky (changes requested):

kotlin-mui-base/src/jsMain/generated/mui/base/Button.classes.kt
turansky: Also - separate file for constants is preferrable

aerialist7: > `external val buttonClasses : ButtonClasses` will work without redundant overrides

Thank you, fixed.

> Also - separate file for constants is preferrable

I believe it's better to place the constant in the same file, the same manner as we do for components (they have their `Props` interfaces in the same file), so hope you're fine with it

kotlin-mui-base/src/jsMain/generated/mui/base/Button.classes.kt
```diff
@@ -17,3 +19,5 @@ external interface ButtonClasses {
     /** State class applied to the root `button` element if `focusVisible={true}`. */
     var focusVisible: ClassName
 }
+
+external val buttonClasses: ButtonClasses
```
turansky, line 23: In new scheme `ButtonClasses` properties must be readonly (`val`s), musn't be?
Also `ButtonClasses` must be `sealed` - inheritance doesn't have big sense.

turansky (changes requested):

kotlin-mui-base/src/jsMain/generated/mui/base/Button.classes.kt
aerialist7: Looks like separate change, isn't it?

turansky: If you update `ButtonClasses` usage, please fix contract too.
Otherwise  `buttonClasses` declaration will be invalid (if it's readonly).

aerialist7: Done

aerialist7:
@Leonya, could you please release these changes?

PR #2274, opened 2024-06-13
TanStack Query `5.45.0` [generated]

porotkin:


kotlin-tanstack-query-core/src/jsMain/generated/tanstack/query/core/resolveStaleTime.kt
```diff
@@ -0,0 +1,10 @@
+// Automatically generated - do not modify!
+
+@file:JsModule("@tanstack/query-core")
+
+package tanstack.query.core
+
+external fun <TQueryFnData, TError, TData, TQueryKey : QueryKey> resolveStaleTime(
+    staleTime: Any, /* undefined | StaleTime<TQueryFnData, TError, TData, TQueryKey> */
```
turansky, line 8: `StaleTime?`

turansky (changes requested):

kotlin-tanstack-query-core/src/jsMain/generated/tanstack/query/core/resolveStaleTime.kt
porotkin: yep, sure!

PR #2300, opened 2024-07-17
Update CHANGELOG.md

xian:
I think this deserves a usage example too?


turansky:
> I think this deserves a usage example too?

Yes, examples before/after will be cool

PR #2304, opened 2024-07-18
Changelog. Add useEffect with suspend callback examples

porotkin:


xian:
That's extremely helpful, thanks so much!

PR #2310, opened 2024-07-20
Enable lock file

sgrishchenko:
We have a long story with `package-lock.json` in Kotlin Wrappers repository:

* In the beginning we wanted to make a contribution more comfortable and we removed `package-lock.json` at all.
* Then we realized that we needed to stabilize builds and we came up with NPM overrides workaround.

But the current state still does not provide enough amounts of stability for builds. In personal discussions often we face strange glitches in `node_modules` so our current strategy for troubleshooting: `gradle clean` + `package-lock.json` removing. It should not be so because it ruins all caches and the whole build stability. And the current approach leads us to "works on my machine" problem.

In this PR I propose to return to `package-lock.json`. From my point of view, it is the most robust approach to provide stable builds. ~~Now NPM support in Kotlin Mutiplatform Gradle plugin becomes more mature. And even NPM itself becomes more mature and can resolve conflicts in `package-lock.json` automatically. Also `package-lock.json` is the most simple and popular approach to stabilize NPM dependencies.~~

~~I tried to describe all common scenarios that can cause problems with `package-lock.json`. I tested these scenarios using my local environment, so probably they should work. If I missed something or you check these approaches and they don't for you, let's discuss them.~~

**UPDATE**
After discussions with @turansky and experiments with `package-lock.json` updating it revealed that manual maintenance of  `package-lock.json` provokes lots of Git conflicts. That is why I tried to change strategy and provide GitHub action that can update `package-lock.json` for us. I tried to describe this approach in an MD file in this PR.

Currently provided GitHub action fully removes exiting `package-lock.json` and generates a new one from scratch. This approach has pros and cons.

Pros:

* We receive information about last publications in transitive dependencies.
* We update transitive dependencies oftener, and we have fewer security vulnerabilities from transitive dependencies.
* We have a simple and fast way to troubleshoot problems from our users. If a user has some runtime problem and bumping of the Kotlin Wrapper version does not help, we suppose that problem might be connected with transitive dependencies, so we ask the user to regenerate their `package-lock.json` and try to reproduce problems. If we have fresh `package-lock.json` in Kotlin Wrapper, it means there is a high probability that versions in Kotlin Wrapper `package-lock.json` and in user's  `package-lock.json` will be similar, and we easily reproduce or detect the root cause of the reported problem.

Cons:

* Every `package-lock.json` update may contain lots of changes, and this can lead to some instability (e.g. in unit tests).
* Because of a significant number of changes after `package-lock.json` update it can be hard to figure out which changes were made because of the version update of our direct dependencies and which changes were made because of new releases of transitive dependencies.


sgrishchenko:
cc @aerialist7 @porotkin

sgrishchenko:
Something went wrong, and the Gradle build failed exactly on  `package-lock.json` check :stuck_out_tongue_winking_eye:. I will move this PR to Draft state until I figure out what is happening.

sgrishchenko:
The problem was with Kotlin Wrappers. It seems that GitHub runs action on the branch after merging. So in the GitHub action environment I have v775 in root `package-lock.json` and v779 in the build directory. It might be a problem to store `package-lock.json` file into Git.

.github/workflows/lock.yml
```diff
@@ -0,0 +1,65 @@
+name: Lock
```
turansky, line 1: About filenames:
* `package-lock.yml`
* `package-lock.md`

looks preferrable

turansky, line 1: name: Package lock?

turansky (changes requested):

turansky (approved):

PR #2324, opened 2024-07-22
React DOM. `useFormState`

porotkin:


porotkin:
it was deprecated in favor of `useActionState` :)

PR #2329, opened 2024-07-27
JS. `Object.fromEntries`

BernhardPosselt:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

### DIFF
--- a/kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
+++ b/kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
@@ -19,6 +19,7 @@ external class Object internal constructor() {
         fun <T : Any> entries(o: ReadonlyRecord<String, T>): ReadonlyArray<JsTuple2<String, T>>
         fun entries(o: Any): ReadonlyArray<JsTuple2<String, Any?>>
         fun <R, T : R> freeze(o: T): R
+        fun <T : Any> fromEntries(entries: ReadonlyArray<JsTuple2<String, T>>): Record<String, T>
         fun <T> getOwnPropertyDescriptor(o: Any, p: PropertyKey): TypedPropertyDescriptor<T>?
         fun getOwnPropertyNames(o: Any): ReadonlyArray<String>
         fun getOwnPropertySymbols(o: Any): ReadonlyArray<Symbol>
--- a/kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
+++ b/kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
@@ -2,6 +2,7 @@ package js.objects
 
 import js.array.JsTuple2
 import js.array.ReadonlyArray
+import js.iterable.JsIterable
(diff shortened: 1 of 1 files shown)


BernhardPosselt:
There might still be something funky with regards to nullability. Technically both methods should also work on nullable objects so both entries and fromEntries should support **T: Any?**

kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
```diff
@@ -19,6 +19,7 @@ external class Object internal constructor() {
         fun <T : Any> entries(o: ReadonlyRecord<String, T>): ReadonlyArray<JsTuple2<String, T>>
         fun entries(o: Any): ReadonlyArray<JsTuple2<String, Any?>>
         fun <R, T : R> freeze(o: T): R
+        fun <T : Any> fromEntries(entries: ReadonlyArray<JsTuple2<String, T>>): Record<String, T>
```
turansky, line 22: 1. Bound for `T` is redundant
2. Return type `Record` -> `ReadonlyRecord`?
3. Could you please add overload with `JsIterable` as parameter?

turansky (changes requested):

kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
BernhardPosselt: pushed

kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
```diff
@@ -19,7 +20,8 @@ external class Object internal constructor() {
         fun <T : Any> entries(o: ReadonlyRecord<String, T>): ReadonlyArray<JsTuple2<String, T>>
         fun entries(o: Any): ReadonlyArray<JsTuple2<String, Any?>>
         fun <R, T : R> freeze(o: T): R
-        fun <T : Any> fromEntries(entries: ReadonlyArray<JsTuple2<String, T>>): Record<String, T>
+        fun <T> fromEntries(entries: ReadonlyArray<JsTuple2<String, T>>): ReadonlyRecord<String, T>
+        fun <T> fromEntries(entries: JsIterable<JsTuple2<String, T>>): ReadonlyRecord<String, T>
```
turansky, line 24: Signature with `JsIterable` - main (`Array: JsIterable`).
Could you move it please on first place?

turansky (changes requested):

kotlin-js/src/jsMain/kotlin/js/objects/Object.kt
turansky: JFYI - `JsIterable` - parent type for `Array`.
Signature with `ReadonlyArray` requied only because parent type missed in Kotlin

turansky (approved):

PR #2330, opened 2024-07-27
MUI. Material `5.16.5` [generated]

aerialist7:


kotlin-mui-material/src/jsMain/generated/mui/material/Portal.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Demos:
+ *
+ * - [Portal](https://mui.com/material-ui/react-portal/)
+ *
+ * API:
+ *
+ * - [Portal API](https://mui.com/material-ui/api/portal/)
+ */
+@JsName("default")
```
turansky, line 19: In previous releases it was removed (`default`)?

aerialist7: No. It was removed in MUI Base, but not in MUI Material

PR #2333, opened 2024-07-27
Allow Object.entries to work on Record<String, Any?>

BernhardPosselt:
I'm not 100% sure because my IDE is still in K2 Alpha mode. I think narrowing down the generic type to T: Any prevents T from working with nullable values in this case, barring you from using it on a Record<String, Any?> 

Please confirm before merging, if this is the correct way to go about it.


turansky:
`<T>` and `<T: Any?>` are equal.
Redundant bound can be removed.

PR #2344, opened 2024-08-02
React Router `6.26.0`. Remix Run Router `1.19.0`

porotkin:


porotkin:
only `.properties` changed. based on #2343

PR #2378, opened 2024-08-10
Tidy Publishing Configuration

aSemy:
- Add `readme.txt` to Javadoc JAR to explain why it's empty.
- Avoid conditional build config - always register the Javadoc JAR task, and avoid 'peeking' into KMP internals to conditionally add it. Always adding it is more robust.
- Use pluginsManager to lazily 'react' to plugins.
- Allow Gradle to set the project's group/artifactId/version in the POM. Manually setting these values is discouraged.
- Add a local-repo - very helpful for debugging.


turansky:

PR #2379, opened 2024-08-10
Create Gradle BuildService for easier aggregation of kotlin-wrappers subprojects

aSemy:
The recent change to automatically aggregate subprojects is nice, but in upcoming Gradle versions they are looking at disabling cross-project configuration, as it harms build configuration performance (https://docs.gradle.org/current/userguide/isolated_projects.html).

Using a BuildService will help with Gradle build performance, and is more future-proof.


turansky:
@aSemy Thank you!

turansky:

PR #2380, opened 2024-08-10
update printBenchmarkResults to register task inputs

aSemy:
Updating Gradle config improves build performance, and prepares for future Gradle versions.

### DIFF
--- a/kotlin-styled-next/build.gradle.kts
+++ b/kotlin-styled-next/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.gradle.api.tasks.PathSensitivity.RELATIVE
+
 plugins {
     `kotlin-legacy-library-conventions`
 }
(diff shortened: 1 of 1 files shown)


kotlin-styled-next/build.gradle.kts
```diff
@@ -33,22 +35,29 @@ dependencies {
...
+    val fileNames = listOf(
+        "AddDuplicateCss",
+        "AddStyledElements",
+        "ConvertToStyled",
+        "CssBuildersInject",
+        "DataTypeOperations",
+        "InjectCssNPlusOne",
+    )
+    val reports = objects.fileCollection().from(fileNames.map { test ->
+        layout.buildDirectory.file("reports/tests/test/classes/benchmark.$test.html")
```
aSemy, line 50: I'm not sure this directory is correct. When I run tests, I can see two, but they're `jsTest` and `jsBrowserTest`

* `kotlin-styled-next/build/reports/tests/jsBrowserTest/`
* `kotlin-styled-next/build/reports/tests/jsTest/`

Do you know if it should really be `reports/tests/test/`?

aSemy:
I also noticed something else while looking at this code.

https://github.com/JetBrains/kotlin-wrappers/blob/299d251ea37debe07903e5d8328caa29a11cf66f/kotlin-styled-next/build.gradle.kts#L85-L91

I don't think `enabled = project.findProperty("test") == true` does anything. I can see the tests do run on CI, when `-Ptest=false`. Is the flag still used? Maybe it should be removed?

kotlin-styled-next/build.gradle.kts
turansky: Probably folder changed after migration on `multiplatform` plugin.

cc @Leonya

turansky:
> I also noticed something else while looking at this code.
> 
> https://github.com/JetBrains/kotlin-wrappers/blob/299d251ea37debe07903e5d8328caa29a11cf66f/kotlin-styled-next/build.gradle.kts#L85-L91
> 
> I don't think `enabled = project.findProperty("test") == true` does anything. I can see the tests do run on CI, when `-Ptest=false`. Is the flag still used? Maybe it should be removed?

Yes, looks like it doesn't work at all :(

PR #2399, opened 2024-08-16
Allow passing parameters to JsClass.newInstance method

BernhardPosselt:


turansky:
@BernhardPosselt What is your use case?

BernhardPosselt:
Creating a new instance of a JS class with parameters. Is there another way to do that?

turansky:
Which library/framework require this action?

BernhardPosselt:
It's a validator/default value API sort of that requires you to instantiate the Schema with the actual object. Not great, but oh well...

BernhardPosselt:
A workaround for that could be wrapping it in a closure instead, but they instanceof check it in other places.

turansky:
API and/or documentation links will be very helpful

BernhardPosselt:
https://foundryvtt.com/api/classes/foundry.abstract.DataModel.html (unfortunately not well documented, had to read 80k lines of JS over the past weeks). Technically I'd need constructor type signatures to make it type safe since those things are instantiated by the framework as well with a specific list of constructor parameters.

What are you looking for specifically? Reflection APIs themselves are usually not type safe and I found it weird that my IDE only completed the no args constructor. Otherwise I'd just close this PR and use Reflect instead.

BernhardPosselt:
On the other hand, I've just taken a look at the Java APIs for this and it looks like the one I was accustomed with (Klass::class.newInstance) was deprecated in favor of a type safe API https://docs.oracle.com/en%2Fjava%2Fjavase%2F11%2Fdocs%2Fapi%2F%2F/java.base/java/lang/Class.html#getDeclaredConstructor(java.lang.Class...). 

In that case the argument for a familiar method might not be as strong as I thought it would have.

turansky:
Did you checked [createInstance](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.full/create-instance.html)?

BernhardPosselt:
I see, so JsClass.newInstance is an analog to this KClass.createInstance. I leave it up to you then. Either way, I can use the Reflect method if required :)

turansky:
`Reflect.construct` - expected **unsafe** solution for now.
For `JsClass.newInstance` update probably more cases required :( 

Updated implementation:
```kotlin
fun <T : Any> JsClass<T>.newInstance(
    vararg parameters: Any?,
): T =
    Reflect.construct(
        target = this,
        argumentsList = parameters,
    )
```

BernhardPosselt:
Thank you for taking your time :)

turansky:
We probably need additional `JsConstructorFunction` for strict constructor signatures (we have some in definitions).
Do you need strict signatures for constructor functions?

BernhardPosselt:
Yes, absolutely. The linked class above as mentioned before requires exactly 2 parameters of a certain type and there are other cases in this project. You can for instance create an https://foundryvtt.com/api/classes/foundry.applications.api.ApplicationV2.html instance but it only works across the board, if it takes a Configuration record subtype as the first parameter.

Personally, I wouldn't design APIs like this but I need to interop with them.

turansky:
Could you please create issue? ;)

Looks like `JsClass.newInstance` will be removed...

BernhardPosselt:
On the Kotlin YouTrack tracker?

turansky:
https://github.com/JetBrains/kotlin-wrappers/issues

turansky:
Or you can just create PR with `JsConstructorFunction` ;)

BernhardPosselt:
For that I'd need to understand how that would be implemented first. In my head I'm thinking of TypeScript constructor interfaces, e.g. 
```ts
type Application = new (config: Configuration) => Application
```
but I don't understand how that would translate to Kotlin nor Kotlin JS

turansky:
Nearest refence - [JsFunction](https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-js/src/jsMain/kotlin/js/function/JsFunction.invoke.kt)

```
// typescript
type Application = new (config: Configuration) => Application

// kotlin
interface Application:
    JsConstructorFunction<JsTuple1<Configuration>, Application>
```

BernhardPosselt:
I think I'm a bit lost since I don't really understand how JsFunction itself works and how it is used.

I think the end goal would be something like:

```kt
fun <P1, C: Any> JsClass<C>.newInstance(
    p1: P1,
): C =
    Reflect.construct(
        target = this,
        argumentsList = arrayOf(p1),
    )
``` 
or maybe:

```kt
inline operator fun <P1, C: Any> JsClass<C>.invoke(
    p1: P1,
): C =
    Reflect.construct(
        target = this,
        argumentsList = arrayOf(p1),
    )
```


Looking at JsFunction itself, it uses a typealias for a context function, so that one obviously can't be overloaded again with another invoke extension.  Interfaces/External interfaces themselves can't define constructor parameters or inherit from classes and you probably want an interface since there's no multiple inheritance.

turansky:
`JsClass` can't be used, because it doesn't contain type parameter for arguments

1. Implementation
```kotlin
@JsExternalInheritorsOnly
external interface JsConstructorFunction<in A : JsTuple, out R>: 
    JsClass<R>
```
2. `invoke` extensions, which will use `Reflect.construct`
3. `adapters` - 1 adapter for start will be fine

PR #2401, opened 2024-08-17
JS. `JsConstructorFunction` (fix #2400)

### LINKED ISSUE #2400: Typesafe interop with JavaScript constructors
I need to interop with JavaScript APIs that expect certain constructors to exist. For instance sub types of https://foundryvtt.com/api/v12/classes/foundry.applications.api.ApplicationV2.html require a constructor that accepts a Configuration record, so that the subclass can be registered as a settings window like this:

```js
game.settings.registerMenu({
    type: ApplicationV2Subclass  // framework creates a new instance using a predefined Configuration object
})
```

Another subclass https://foundryvtt.com/api/v12/classes/foundry.applications.api.DocumentSheetV2.html requires a Configuration record with an additional document parameter (a Fat Model to handle Database I/O). 

All of these classes are also available as properties on these subclasses so that you can create new instances if you need to. Right now, there is no way to define that sub classes need to have a certain constructor, nor a way to define that a property on a JS class must have that signature and can be instantiated with a certain constructor call

Implementation discussions - #2399

BernhardPosselt:


kotlin-js/src/jsMain/kotlin/js/function/JsConstructorFunction.adapter.kt
```diff
@@ -0,0 +1,8 @@
+package js.function
+
+import js.array.*
+
+fun <C: JsClass<C>> JsConstructorFunction(
+    fn: C.() -> C,
```
turansky, line 6: For start it can be unsafe.
Single parameter - `JsClass`

BernhardPosselt:
Wanted to ask in the original thread and opened this one by accident on the wrappers repo.

The invoke defs are easy enough, but there are 2 things that are unclear for me:

* out Variance for R is not allowed in that position since R in JsClass is invariant. Could also be my IDE which right now has a ton of parser crashes making this extra difficult. Not sure how to solve that especially because variance here looks sensible with regards to PECS. Require use site variance for R instead?
    ```js
    @JsExternalInheritorsOnly
    external interface JsConstructorFunction<in A : JsTuple, R>: 
    JsClass<R>
    ```
* Not sure what kind of issue the adapter should solve here or what it solves exactly for JsContextFunction.

kotlin-js/src/jsMain/kotlin/js/function/JsConstructorFunction.adapter.kt
BernhardPosselt:
```kt
fun <C: JsClass<C>> JsConstructorFunction(
    clazz: JsClass<C>,
): JsConstructorFunction<JsTuple, C> = clazz.unsafeCast<JsConstructorFunction<JsTuple, C>>()
```
like this?

turansky:
1. `out R` -> `R` - fine change
2. Adapter should "convert" `JsClass` to `JsConstructorFunction`
    * For start it will be unsafe and will expect, that default constructor of class is fine

kotlin-js/src/jsMain/kotlin/js/function/JsConstructorFunction.adapter.kt
```diff
@@ -2,7 +2,12 @@ package js.function
 
 import js.array.*
 
-fun <C: JsClass<C>> JsConstructorFunction(
-    fn: C.() -> C,
-): JsConstructorFunction<JsTuple, C> = TODO()
+fun <C: Any> JsConstructorFunction(
```
turansky, line 5: Arguments expected in type parameters

turansky (changes requested):

kotlin-js/src/jsMain/kotlin/js/function/JsConstructorFunction.adapter.kt
turansky: Type parameters should be the same like in `JsConstructorFunction` declaration.
Same order, same names.

BernhardPosselt: Like this?
```kt
fun <A : JsTuple, R : Any> JsConstructorFunction(
    clazz: JsClass<R>,
): JsConstructorFunction<A, R> = clazz.unsafeCast<JsConstructorFunction<A, R>>()
```

turansky: Yes

kotlin-js/src/jsTest/kotlin/js/function/JsConstructorFunctionTest.kt
```diff
@@ -0,0 +1,20 @@
+package js.function
+
+import js.array.JsTuple
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@OptIn(ExperimentalJsExport::class)
```
turansky, line 7: Could you please enable it by default [here](https://github.com/JetBrains/kotlin-wrappers/blob/3148da8ee3b193aa2a31275386c86f0072317aef/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts#L39)?

turansky (changes requested):

turansky:
@BernhardPosselt Thank you!

turansky:

PR #2431, opened 2024-08-26
React Types `18.3.4`

porotkin:


kotlin-react-dom/src/jsMain/generated/react/dom/DOMAttributes.kt
```diff
@@ -56,6 +56,12 @@ external interface DOMAttributes<T : Element> :
     // Keyboard Events
     var onKeyDown: KeyboardEventHandler<T>?
     var onKeyDownCapture: KeyboardEventHandler<T>?
+
+    /** @deprecated Use `onKeyUp` or `onKeyDown` instead */
```
turansky, line 60: We don't declare deprecated members normally

turansky (changes requested):

PR #2512, opened 2024-09-15
Add JsPlainObject to MakeDirectoryOptions

Vampire:


turansky (changes requested):
Child interfaces also must be marked with `@JsPlainObject`

turansky (approved):

PR #2522, opened 2024-09-18
Add wrappers for semver

Vampire:


kotlin-semver/src/jsMain/generated/semver/functions/inc/IdentifierBase.kt
```diff
@@ -0,0 +1,15 @@
+// Generated by Karakum - do not modify it manually!
+
+package semver.functions.inc
```
turansky, line 3: `inc` looks like redundant subpackage

kotlin-semver/src/jsMain/generated/semver/RangeOptions.kt
```diff
@@ -0,0 +1,10 @@
+// Generated by Karakum - do not modify it manually!
+
+package semver
+
+import semver.classes.SemVer
+
+@kotlinx.js.JsPlainObject
+external interface RangeOptions : Options {
+var includePrerelease: Boolean?
```
turansky, line 9: Could you please format generated sources?

kotlin-semver/src/jsMain/generated/semver/functions/sort.kt
```diff
@@ -0,0 +1,28 @@
...
+import semver.classes.SemVer
+
+/* import SemVer = require("../classes/semver"); */
+
+/* import semver = require("../index"); */
+
+/**
+ * Sorts an array of semver entries in ascending order using `compareBuild()`.
+ */
+external fun <T : Any /* string | SemVer */> sort(list: Array<T>): Array<T>
```
turansky, line 16: Could you please report bug on Karakum with cases where `Array` generated?
In all of them `ReadonlyArray` expected :(

turansky (changes requested):

kotlin-semver/src/jsMain/generated/semver/functions/inc/IdentifierBase.kt
Vampire: Well, it is the package-strategy `inc` namespace you two agreed on Discord is the best solution for the case.
I guess it is intended to have separate namespace from the other functions in case they als need some classes or typedefs. Do you think differently now how it should be solved?

kotlin-semver/src/jsMain/generated/semver/RangeOptions.kt
Vampire: I would be delighted to, the strange readability constantly bugged me.
But I thought the generated sources should not be touched?
Also, if you want the generated sources to be uniformly and nicely formatted, maybe after generation some automatica formatting should be happening, for example by sending the files through ktlint?

kotlin-semver/src/jsMain/generated/semver/functions/sort.kt
Vampire: Sure, how about `Error`?
Those are not here but when I prepare PR for `null-writable` where I used a plugin to change them to `js.errors.JsError`.

kotlin-semver/src/jsMain/generated/semver/RangeOptions.kt
turansky: > maybe after generation some automatica formatting should be happening, for example by sending the files through ktlint?

Automatic tool - under discussion.
For now - formatting + cleanup in IDEA is what we use.

kotlin-semver/src/jsMain/generated/semver/functions/sort.kt
Vampire: I created a PR for Karakum at https://github.com/karakum-team/karakum/pull/28 and for now added a plugin here that changes the types.

kotlin-semver/src/jsMain/generated/semver/functions/inc/IdentifierBase.kt
turansky: It's my fault :(
I thought, that redundand package won't be created

turansky:
@Vampire thank you for contribution!

3 additional thinks, which can be fixed:

1. Missed trailing commas
    * We have it in `.editorconfig`, probably you should reapply it in IDEA
2. `devNpm` to types dependency
3. Naming for operators - [GPUCompareFunction](https://github.com/JetBrains/kotlin-wrappers/blob/fe5cb872ad66c097c720d53ae217538d9cf57ea8/kotlin-web/src/jsMain/generated/web/gpu/GPUCompareFunction.kt#L22)

Vampire:
If you tell me before merging, I can fix it before merging. :-D
But here you have it: #2526 :-)

I don't know what was the issue with the uglification of point 1.
Sometimes IJ did add them, sometimes not.
I did now send ktlint over it additionally, now they are there, and it also matches better the "one parameter per line" style you mentioned.

turansky:

PR #2528, opened 2024-09-18
Fix SemVer module declarations

Vampire:


turansky:
Probably in this case `ncc` and bundlers will include whole library inside bundle if any function used.

Vampire:
> Probably in this case ncc and bundlers will include whole library inside bundle if any function used.

That's correct if you don't use "minify".
So here with this change it is 8.3 MiB instead of 8.2 MiB with the "proper fix" of moving the annotation from the file to the class.
While the "proper fix" does not work properly if the consuming project is an ESM.
Ncc packing works (did not try to run the result) but running the result before NCC packing did not work.

PR #2529, opened 2024-09-19
Add wrappers for null-writable

Vampire:


kotlin-null-writable/src/jsMain/generated/nullwritable/NullWritable.kt
```diff
@@ -0,0 +1,24 @@
...
+
+@file:JsModule("null-writable")
+
+package nullwritable
+
+// unhandled import: Writable from "node:stream"
+import node.buffer.BufferEncoding
+import node.stream.Writable
+
+external class NullWritable : Writable {
```
turansky, line 11: Looks like we can just add implementation from original (2 lines of code) and remove Karakum generation :)

Vampire: Yes, if you prefer, we can do it.
I just thought where possible, generation is preferable.

Vampire: I even made a PR to that project first to bring the declaration in line with upstream declaration, so that the generation immediately generates the proper declarations that can `override` :-D

turansky: > Yes, if you prefer, we can do it.

Yes, please.
Especially when we know about open problem with NCC packaging.

Vampire: I don't get what you are saying.
The NCC packaging problem has nothing to do with whether the sources are generated or manually checked in.
I tried it in a pure plain old TS project and it happened there just the same.

PR #2711, opened 2024-12-25
MUIX. Date Pickers `7.23.2` [generated]

aerialist7:


kotlin-muix-date-pickers/src/jsMain/generated/muix/pickers/DesktopDatePicker.types.kt
```diff
@@ -0,0 +1,27 @@
...
+     * The props used for each component slot.
+     * @default {}
+     */
+    var slotProps: DesktopDatePickerSlotProps<TDate, TEnableAccessibleFieldDOMStructure>?
+
+    /**
+     * Years rendered per row.
+     * @default 4
+     */
+    var yearsPerRow: dynamic
```
turansky, line 22: Please use `Any` or `Any?` instead here and in other declarations.
`dynamic` type usage is strictly not recommended as non-multiplatform.

turansky (changes requested):

PR #2730, opened 2025-01-20
Add wrappers for yaml

Vampire:


kotlin-yaml/src/jsMain/kotlin/yaml/parse.kt
```diff
@@ -0,0 +1,19 @@
...
+     * support you should use `YAML.parseAllDocuments`. May throw on error, and may
+     * log warnings using `console.warn`.
+     *
+     * @param str - A string with YAML formatting.
+     * @param reviver - A reviver function, as in `JSON.parse()`
+     * @returns The value will match the type of the root value of the parsed YAML
+     *   document, so Maps become objects, Sequences arrays, and scalars result in
+     *   nulls, booleans, numbers and strings.
+     */
+    fun <T> parse(src: String): T
```
turansky, line 18: Top-level function expected. `YAML` object is redundant

turansky (changes requested):

kotlin-yaml/src/jsMain/kotlin/yaml/parse.kt
turansky, line 18: Non-null `T`?

turansky (changes requested):

kotlin-yaml/src/jsMain/kotlin/yaml/parse.kt
Vampire: My intention was to have it similar to the typical TS usage which is `YAML.parse` similar to `JSON.parse`. But I can of course change it.

Vampire: `T` can be a nullable type, so does it matter?

Vampire:
Btw. you should not merge this before #2729, or build will probably fail due to #2728.
At least it did so for me after I found out why the new module is not appearing in the catalog and added that check.

turansky (changes requested):

settings.gradle.kts
```diff
@@ -203,3 +203,6 @@ include("kotlin-web")
 include("kotlin-wrappers-bom")
 
 include("kotlin-wrappers-catalog")
+
+// Kotlin/JS: yaml wrappers
+include("kotlin-yaml")
```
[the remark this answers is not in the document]
Vampire: Because you want it before, or because you think it would fix the ordering problem?
The latter is not the case.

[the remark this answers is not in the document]
Vampire: Actually I wonder why the build fails right now, as I reverted the rebase onto master. :-/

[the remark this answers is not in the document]
turansky: Structure - main reason

- declarations
- bom
- catalog

kotlin-yaml/src/jsMain/kotlin/yaml/parse.kt
turansky: Intention is valid.
Screenshot 2025-01-20 at 23 25 29
In both cases `parse` looks like "common" name. Other names are more unique and don't need namespace.
If it will be a problem we will use more specific names.
In case of `yaml` it's also bundle size problem :(

Vampire: You can also do `yaml.parse(...)` instead of an import, so maybe that is also good enough if it is more consistent for the wrappers codebase.

What do you mean by bundle size?
Is the bundle bigger if you use an object namespace?
Or what did you mean?

Vampire: Hm, actually I think `Any?` might be better, otherwise you have to type the call.
With `T` I cannot just do `yaml.parse("null")` or `val foo = yaml.parse("null")` but have to do `val foo: Any? = yaml.parse("null")`.
On the other hand with `Any?` then have to cast if you do want to assign it somewhere typed.
What do you think?

turansky: > Is the bundle bigger if you use an object namespace?

Yes, all object will be included

turansky: [Done](https://github.com/JetBrains/kotlin-wrappers/commit/2ed2b48897c283188ffbd703aa2b061e78fdc945)

Vampire: Ah, ok, judging by `JSON.parse` it seems you prefer it like this with `T` :-)

Vampire: Oh, I didn't even mean in kotlin wrappers, but in TS or JS. :-D
But yeah, within this codebase it should probably also be consistent and that's more important. :-)
Now I just wonder that you didn't pull out the other JSON functions. :-D

Vampire: Oh, you did. :-D

turansky: `JSON.parse` also can return nullable values :(

Vampire: Yes, indeed interestingly it just works even with the old signature.
But I guess removing the `: Any` would be appropriate.

PR #2733, opened 2025-01-23
Use Kotlin DSL for GHA workflow files

Vampire:
As you like Kotlin and also there was some repetition in your workflow scripts,
I thought you might like using the Kotlin DSL `github-workflows-kt` provides to define
your GitHub action workflows. :-)

Once https://github.com/karakum-team/vitalik/pull/1 was considered I'll update
this PR to use the typed `Vitalik` class instead of the untyped one.
But I thought I send the PR already because other than that, it would be ready.


turansky:
Looks like we will lose existed IDE support GitHub actions after this PR :(

Vampire:
I wouldn't say "loose", as there is no real IDE support.
There is some minimal schema support if the schemastore is used, but it also ist not too intelligent.

The IDE support with the Kotlin DSL is actually much bigger and better.
And it allows DRY which the YAML files do not support as anchors and aliases are not enabled for those.

Due to applying DRY principle by having the common stuff in `common.main.kts`, you indeed trigger an IntelliJ bug ([caveat #2 in the readme](https://youtrack.jetbrains.com/issue/KTIJ-14580)) that currently breaks the IDE support.

But this is a JetBrains repository, maybe you have the right people to contact to finally get this annoying issue fixed, as was announced in the blog post some while ago. :-)

To mitigate the caveat you can just copy over the `DependsOn`s and their `Repository`s from the common script (and maybe the `buildJob` signature while editing the script and then remove them again. The `DependsOn`s and `Repository`s could even stay in the main scripts, that would just not be as DRY, then for editing only the `buildJob` signature needs to be copied over.

Another way if that is not an option would be to loosen the DRY and not put the common stuff to `common.main.kts` but have each main script self-contained until KTIJ-14580 is finally fixed.

PR #2742, opened 2025-02-15
Kotlin-based Karakum integration for Node declarations

sgrishchenko:
In this PR I would like to demonstrate how I see integration with a new version of Karakum, that is written in Kotlin now.

Note: this PR is in WIP status because additional preparations from Karakum side are needed.


buildSrc/src/main/kotlin/karakum-conventions.gradle.kts
```diff
@@ -0,0 +1,13 @@
+import org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsExec
+
+plugins {
+    id("kotlin-conventions")
```
turansky, line 4: Please copy required configuration instead

turansky (changes requested):

buildSrc/src/main/kotlin/karakum-conventions.gradle.kts
sgrishchenko: Fixed

PR #2758, opened 2025-03-19
React Select: Add `Value.isNotEmpty` helper

aerialist7:


turansky (commented):
1. NPE will be received in `null` case
2. @sgrishchenko planned to provide common version of this interface

PR #2775, opened 2025-04-05
JS. Add missing methods and extension functions to `JsNumber` class.

rjaros:


turansky:
@rjaros do you have use cases for those methods?

rjaros:
Yes. I'm using `toFixed` and `toLocaleString` in my project. I've added `toPrecision` just for completeness.

kotlin-js/src/commonMain/kotlin/js/core/JsNumber.kt
```diff
@@ -10,6 +12,18 @@ external class JsNumber {
     internal constructor(value: String)
 
     fun toExponential(fractionDigits: Int = definedExternally): String
+    fun toFixed(digits: Int = definedExternally): String
+    fun toLocaleString(
```
turansky, line 16: Please use `Localizable` interface instead

kotlin-js/src/commonMain/kotlin/js/core/JsNumber.kt
```diff
@@ -33,3 +47,57 @@ fun Double.toExponential(): String =
...
+fun Int.toLocaleString(locales: String): String =
+    JsNumber(this).toLocaleString(locales)
+
+fun Int.toLocaleString(locales: String, options: NumberFormatOptions): String =
+    JsNumber(this).toLocaleString(locales, options)
+
+fun Int.toLocaleString(locales: ReadonlyArray<JsString>): String =
+    JsNumber(this).toLocaleString(locales)
+
+fun Int.toLocaleString(locales: ReadonlyArray<JsString>, options: NumberFormatOptions): String =
```
turansky, line 75: Please use similar formatting as for methods (2+ parameters = on separate lines)

turansky (changes requested):

rjaros:
Is it ok to use externals without explicit `JsAny` interface? It results in errors in IDE.

turansky:
> It results in errors in IDE.

Looks like IDE bug ;)

PR #2786, opened 2025-05-15
Build. Update `karakum-legacy` generator tasks to comply with configuration caches

porotkin:


generate-logic-legacy/src/main/kotlin/Directories.kt
```diff
@@ -0,0 +1,8 @@
+import org.gradle.api.file.Directory
+import org.gradle.api.file.DirectoryProperty
+
+val DirectoryProperty.rootGeneratedDir: Directory
+    get() = dir("generated").get()
```
sgrishchenko, line 5: It is a bit strange to unwrap a property to its value forcely. Most gradle configs accept both `T | Property<T>`.

generate-logic-legacy/src/main/kotlin/GenerateDeclarations.kt
```diff
@@ -0,0 +1,64 @@
...
+
+abstract class GenerateDeclarationsTask
+@Inject constructor(
+    private val projectLayout: ProjectLayout,
+    private val fileSystemOperations: FileSystemOperations,
+    objectFactory: ObjectFactory,
+) : DefaultTask(), HasGeneratedDirectories {
+
+    private val rootBuildDirectory: Directory
+        get() = project.rootProject.layout.buildDirectory.get()
```
sgrishchenko, line 27: In my understanding, using of Properties is more preferable because Gradle supports different mechanics that required lazy evaluation. E.g. if you want to customize some task properties, your customization can have no effect because you already read the non-property task parameter in the task constructor.

generate-logic-legacy/src/main/kotlin/declarations.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
     id("io.github.turansky.kfc.library")
 }
 
-val generateDeclarations by tasks.registering {
+val generateDeclarations by tasks.registering(GenerateDeclarationsTask::class) {
```
sgrishchenko, line 6: To be honest, I don't understand what is happening here. We are creating some tasks without any parametrization, that just install npm deps for us. We don't configure nether `sourceDirs` nor `action` here, so what the point to use `GenerateDeclarationsTask` here?

sgrishchenko (approved):

generate-logic-legacy/src/main/kotlin/declarations.gradle.kts
porotkin: @sgrishchenko, we actually configure the created tasks in other convention plugins

turansky (changes requested):
More elegant solution expected

sgrishchenko:
> More elegant solution expected

In my understanding solving of "do <something> to comply <something>" tasks suppose just minimal efforts to achieve some green mark (successful compilation or successful CI run). If we want to provide some elegant solutions, it is about "refactore/improve <something>" tasks.

And to be honest, it is unclear what is required to achieve "elegant solution".

turansky:
> And to be honest, it is unclear what is required to achieve "elegant solution".
 
 @sgrishchenko it means "do nothing" or "do less changes"

sgrishchenko:
@turansky Funny, I got it in completely opposite meaning :slightly_smiling_face: I read it like "You made some efforts, please, do more (to support code in the folder with the "legacy" word in the name)". Now I hope I got it, thanks.

PR #2788, opened 2025-06-05
CSS. Add `wasmJs` target

VedaVoronina:


kotlin-css/build.gradle.kts
```diff
@@ -13,6 +16,8 @@ kotlin {
     iosSimulatorArm64()
     iosX64()
 
+    wasmJs()
```
turansky, line 19: After `js()`?

turansky:

Issue #2449, opened 2024-08-31
Add `wasmJs` target for `kotlin-browser`


turansky:
Open problems:
* No common `definedExternally`

Required `expected` types:
* `JsAny`
* `JsBoolean`
* `JsNumber`
* `JsString`
* `JsArray`
* `JsError`

Check:
* `ReadonlyArray`
* `JsLong`

Kotlin issue - TBD

[Related discussion](https://kotlinlang.slack.com/archives/C0B8L3U69/p1724961976083689)

turansky:
Initial support implemented

ApoloApps:
will this replace kolinx browser library?? Since that one has been updated in months.....

PR #2795, opened 2025-06-27
Web. Remove `@JsValue` from `Ed25519` (#2449)

zibet27:
Remove the remaining JsValue annotations from `browser` and `web` packages. [https://github.com/JetBrains/kotlin-wrappers/issues/2449](https://github.com/JetBrains/kotlin-wrappers/issues/2449)


kotlin-browser/src/commonTest/kotlin/web/svg/SVGTest.kt
```diff
@@ -0,0 +1,10 @@
+package web.mathml.svg
```
turansky, line 1: Strange package

turansky (changes requested):

PR #2834, opened 2026-01-30
 JS. Add extension functions for converting maps, sequences, iterables into records

BernhardPosselt:
I wasn't sure where to put those extension functions. If you can think of a better location let me know.

With regards to creating maps from iterables and sequences, there is no guarantee to that multiple keys aren't the same. Right now, the last value in the chain is being kept (as implemented in Object.fromEntries), but if you want to add a version with a [merge function](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-java.util.function.BinaryOperator-) let me know.

Lastly I'm unsure if we have a way to test for data equality with regards to Records, so I went with JSON.stringify which is common in JS

### DIFF
--- a/kotlin-js/src/webMain/kotlin/js/objects/ReadonlyRecord.kt
+++ b/kotlin-js/src/webMain/kotlin/js/objects/ReadonlyRecord.kt
@@ -24,23 +24,23 @@ inline operator fun <V : JsAny?> ReadonlyRecord<JsAny, V>.get(key: String): V? =
 inline operator fun <V : JsAny?> ReadonlyRecord<JsString, V>.get(key: String): V? =
     get(key.toJsString())
 
-fun <V : JsAny?> Sequence<Tuple2<JsString, V>>.toReadonlyRecord(): ReadonlyRecord<JsString, V> =
+fun <K: JsAny, V : JsAny?> Sequence<Tuple2<K, V>>.toReadonlyRecord(): ReadonlyRecord<K, V> =
     Record {
         forEach { (key, value) -> set(key, value) }
     }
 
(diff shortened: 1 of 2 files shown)


BernhardPosselt:
Do you know where toTypedArray is located?

image

My guess is that Collection does not implement the extension function in this compilation target?

BernhardPosselt:
And is this the way to write tests? When I move the test into the webTest target, I'm losing access to Kotlin's default types:

```kt
val expected = Record {
            set("one".toJsString(), 1.toJsNumber())
            set("two".toJsString(), 2.toJsNumber())
        }
        val actual = sequenceOf(
            "one".toJsString() to 1.toJsNumber(),
            "two".toJsString() to 2.toJsNumber(),
        ).toRecord()
```

BernhardPosselt:
Tests took considerable effort to convert since many conversion methods were not available. I hope I managed to do it properly. Funnily enough the actual implementation could be replaced by the same piece of code in all cases. I can do another copy of test cases if you are happy with the test itself.

BernhardPosselt:
Ah, you mean:

```kt
fun <V : JsAny> Map<String, V>.toRecord(): Record<JsString, V> =
    Record {
        forEach { (key, value) -> set(key.toJsString(), value) }
    }
 ```

**Comment by turansky** [comment]
`K` parameter
```kotlin
fun <K: JsAny, V : JsAny?> Map<K, V>.toRecord(): Record<K, V> =
    Record {
        forEach { (key, value) -> set(key, value) }
    }
```

Also please align `V` bound - it should be same as in records - `V : JsAny?` (nullable)

PR #2849, opened 2026-03-26
TanStack React Router. Support `useMatch`

aerialist7:


turansky (changes requested):

turansky (changes requested):

kotlin-tanstack-react-router/src/jsMain/kotlin/tanstack/react/router/useMatch.ext.kt
```diff
@@ -0,0 +1,18 @@
+package tanstack.react.router
+
+import tanstack.router.core.RouteId
+import tanstack.router.core.RouteMatch
+
+fun useMatch(): RouteMatch =
```
turansky: Missed `UseMatchResult` alias?

turansky (changes requested):

PR #2850, opened 2026-03-26
TanStack React Router. Support `useMatches`

aerialist7:


turansky (changes requested):

turansky (changes requested):

turansky (changes requested):

turansky (changes requested):

sgrishchenko:
I will try to fix the misunderstanding. There are a finite number of patterns that are good for TS code, and bad or not so good for Kotlin. Some examples:

* Callback last: 
  * Bad: `useEffect({ ... }, deps)` 
  * Good: `useEffect(deps) { ... }` 

* Options vs Builder:
  * Bad: `useSomeConfig(SomeConfig(one = 1, two = 2))`
  * Better: `useSomeConfig { one = 1, two = 2 }`

Maybe there are more, but I can't remember now (maybe it is time to list them...).

Anyway, if you meet some of these patterns, it is worth considering hiding the original declaration as raw and decorating it with a wrapper. But if the original TS declaration is OK and does not meet any of those patterns, there is no much sense in decoration. 

Different libraries have different lists of patters, so I think a good start is declaring signature without wrapping at all and try to use it somewhere (same file with `main` function, neighboring file, test file, example...). If usage is comfortable, I think you can send PR with this declaration. If usage is uncomfortable, you can inspect similar declarations, or you can create PR with the description: "Guys, I want something like this, but I don't understand how to make this declaration better, can you point to something similar and existing or consider this case if it is unique?"

PR #2851, opened 2026-04-16
Floating UI DOM. `autoUpdate`

porotkin:
[Docs](https://floating-ui.com/docs/autoupdate)

### DIFF
--- a/kotlin-floating-ui-dom/src/webMain/kotlin/floating/ui/dom/AutoUpdateOptions.kt
+++ b/kotlin-floating-ui-dom/src/webMain/kotlin/floating/ui/dom/AutoUpdateOptions.kt
@@ -1,0 +1,40 @@
+package floating.ui.dom
+
+import kotlinx.js.JsPlainObject
+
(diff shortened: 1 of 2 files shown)


kotlin-floating-ui-dom/src/webMain/kotlin/floating/ui/dom/FloatingElement.kt
```diff
@@ -0,0 +1,5 @@
+package floating.ui.dom
+
+import web.html.HTMLElement
+
+typealias FloatingElement = HTMLElement
```
turansky: Please check if it's required alias

turansky (changes requested):

kotlin-floating-ui-dom/src/webMain/kotlin/floating/ui/dom/FloatingElement.kt
porotkin: reverted the unnecessary alias

PR #2853, opened 2026-04-24
fix: Exception during run: SyntaxError[ ...  ]: Identifier 'set' has already been declared for csstypes

mkienenb:
Fixes #2852  (example proof)

Also fixes my own project.


mkienenb:
Closing -- cannot update Record.

PR #2856, opened 2026-05-31
kotlin-pako: Add streaming `Deflate` and `Inflate` class declarations

MohammedKHC:
- Inflate: https://github.com/nodeca/pako/blob/master/lib/inflate.js#L60-L169
- Deflate: https://github.com/nodeca/pako/blob/master/lib/deflate.js#L64-L176


MohammedKHC:
@turansky
NOTE: I briefly adapted the documentation to better match Kotlin conventions.

MohammedKHC:
I'm also unsure whether `push()` should accept `Uint8Array<ArrayBuffer>` or `Uint8Array<*>` — since callers may have Uint8Array backed by different buffer types, Uint8Array<*> feels more correct and flexible.

So what do you think?

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
```diff
@@ -0,0 +1,57 @@
...
+     * new output chunks. Returns `true` on success. If end of stream detected,
+     * [onEnd] will be called.
+     *
+     * `flush_mode` is not needed for normal operation, because end of stream
+     * detected automatically. You may try to use it for advanced things, but
+     * this functionality was not tested.
+     *
+     * On fail call [onEnd] with error code and return false.
+     */
+    fun push(data: Uint8Array<*>, flush: Flush): Boolean
```
turansky: I see 3 overloads and optional mode [here](https://nodeca.github.io/pako/#Deflate.prototype.push).

turansky (changes requested):

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import js.buffer.ArrayBuffer
+import js.typedarrays.Uint8Array
+
+/**
+ * Generic JS-style wrapper for zlib calls. If you don't need
+ * streaming behaviour - use more simple functions: [inflate]
+ * and [inflateRaw].
+ */
+external class Inflate(options: InflateOptions = definedExternally) {
```
turansky: Is it open?

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Error code after inflate finished. [ReturnCode.Z_OK] on success.
+     * Should be checked if broken data possible.
+     */
+    val err: ReturnCode
+
+    /**
+     * Error message, if [err] != ReturnCode.Z_OK
+     */
+    val msg: String
```
turansky: Nullable?

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
```diff
@@ -0,0 +1,57 @@
...
+     * and [onEnd] handlers. Filled after you push last chunk
+     * (call [push] with [Flush.Z_FINISH] param).
+     */
+    val result: Uint8Array<ArrayBuffer>
+
+    /**
+     * Error code after inflate finished. [ReturnCode.Z_OK] on success.
+     * Should be checked if broken data possible.
+     */
+    val err: ReturnCode
```
turansky: Nullable?

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
```diff
@@ -0,0 +1,57 @@
...
+ * streaming behaviour - use more simple functions: [inflate]
+ * and [inflateRaw].
+ */
+external class Inflate(options: InflateOptions = definedExternally) {
+    /**
+     * Uncompressed result, generated by default [onData]
+     * and [onEnd] handlers. Filled after you push last chunk
+     * (call [push] with [Flush.Z_FINISH] param).
+     */
+    val result: Uint8Array<ArrayBuffer>
```
turansky: Is it nullable at start?

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
MohammedKHC: I think that err shouldn't be nullable as it just defaults to ReturnCode.Z_OK which is 0?
What do you think?

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
MohammedKHC: Yes! I forget this

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
MohammedKHC: Hmmm probably yes

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
MohammedKHC: So in my opinion it should not be open. Because it's so convenient to just set the onData function.

However if we were to make it open then we should replace the onData/onEnd with functions.
Iam not 100% sure what is the right thing to do

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
turansky: I want real type ;)
If type is valid - problem is solved.

kotlin-pako/src/webMain/kotlin/pako/Inflate.kt
turansky: > However if we were to make it open then we should replace the `onData/onEnd` with functions.

Yes, it looks more kotlinish.
If it works - it's preferrable option.

kotlin-pako/src/webMain/kotlin/pako/Deflate.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import js.buffer.ArrayBuffer
+import js.typedarrays.Uint8Array
+
+/**
+ * Generic JS-style wrapper for zlib calls. If you don't need
+ * streaming behaviour - use more simple functions: [deflate],
+ * [deflateRaw] and [gzip].
+ */
+external class Deflate(options: DeflateOptions = definedExternally) {
```
turansky: Please put parameters on separate lines with trailing comma.
Here and in methods

turansky (changes requested):

turansky (changes requested):
Please add `.class` suffix to file names (`Deflate.class.kt` for example) to avoid file name conflicts on non-Linux operation systems.

MohammedKHC:
@turansky I applied the requested changes.
What do you think?

turansky:

turansky:
@MohammedKHC thank you!

MohammedKHC:
Thank you for your responsiveness, you are literally the fastest maintainer I have ever worked with!

PR #2859, opened 2026-06-09
Cesium. Fix `Appearance` hierarchy

porotkin:


turansky (changes requested):
Nearest context isn't analyzed

PR #2861, opened 2026-06-26
Cesium. Add `open` modifier for overridden members in open classes

porotkin:
Resolves the `all open?` todo


turansky:
Describe change has no sense from Kotlin perspective :(

PR #2872, opened 2026-08-05
Use jsName for external interfaces with companion object

alex28sh:
In the scope of "[KT-76462](https://youtrack.jetbrains.com/projects/KT/issues/KT-76462) K/Wasm: Redundant `NESTED_CLASS_IN_EXTERNAL_INTERFACE` error" external interfaces with companion object should start using JsName("globalThis") or JsName("Object") (until companion blocks will be stabilized)

I found globalThis more appropriate: these objects should not support reflection and with globalThis TypeError will be thrown in case of attempt to check isinstance. With Object it will just return true every time.

However, the following check succeeds (in both cases)
```CanvasFillRule.Companion == OtherExternalInterface.Companion```


alex28sh:
@sgrishchenko Could you review, please?

I regenerated kotlin-js, kotlin-web with `./gradlew -p kotlin-browser/karakum syncKotlinJs syncKotlinWeb` - is it how it should be?

Should related changes be applied to karakum?

sgrishchenko:
@alex28sh Thank you for your changes! Generating of kotlin-js and kotlin-web mostly managed by @turansky, I requested him to review it. I will also support it in the karakum and regenerate projects.

alex28sh:
Updated with `@JsName("Object")`, according to our current recommendations for Kotlin/Wasm users

turansky:
@alex28sh thank you!

PR #2873, opened 2026-08-11
CSS. Add `clip` to `Overflow` enum

CC007:
See https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow#values for an explanation of what the clip value does


turansky:
@CC007 thank you!

PR #2874, opened 2026-08-12
Add AxialInset data class and related properties for inset handling

CC007:
This adds support for `inset-inline`, `inset-block`, `inset-inline-start`, `inset-inline-end`, `inset-block-start` and `inset-block-end`

Docs for the CSS elements:
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-start
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-end
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block-start
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block-end

### DIFF
--- a/kotlin-css/src/commonMain/kotlin/kotlinx/css/Inset.kt
+++ b/kotlin-css/src/commonMain/kotlin/kotlinx/css/Inset.kt
@@ -26,7 +26,14 @@ data class Inset(
     )
 }
 
-data class LinearInset(
+/**
+ * Inset for a specific axis.
+ * Axial insets are affected by the specified writing-mode and direction.
+ *
+ * @param start the inset at the start of the axis
+ * @param end the inset at the end of the axis
+ */
+data class AxialInset(
     val start: LinearDimension,
     val end: LinearDimension,
 ) : CssValue(
@@ -38,8 +45,8 @@ data class LinearInset(
 }
 
 @get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.inset: Inset by CssProperty()
-@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetInline: LinearInset by CssProperty()
-@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetBlock: LinearInset by CssProperty()
+@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetInline: AxialInset by CssProperty()
+@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetBlock: AxialInset by CssProperty()
 @get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetInlineStart: LinearDimension by CssProperty()
 @get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetInlineEnd: LinearDimension by CssProperty()
 @get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetBlockStart: LinearDimension by CssProperty()
--- a/kotlin-css/src/commonMain/kotlin/kotlinx/css/Inset.kt
+++ b/kotlin-css/src/commonMain/kotlin/kotlinx/css/Inset.kt
@@ -26,4 +26,21 @@ data class Inset(
     )
 }
 
+data class LinearInset(
+    val start: LinearDimension,
+    val end: LinearDimension,
+) : CssValue(
+    getShorthandValue(start, end)
+) {
+    override fun toString() = value
+
+    constructor(all: LinearDimension) : this(all, all)
+}
+
 @get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.inset: Inset by CssProperty()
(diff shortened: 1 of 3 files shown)


kotlin-css/src/commonMain/kotlin/kotlinx/css/Inset.kt
```diff
@@ -26,4 +26,28 @@ data class Inset(
...
 }
 
+/**
+ * Inset for a specific axis.
+ * Axial insets are affected by the specified writing-mode and direction.
+ *
+ * @param start the inset at the start of the axis
+ * @param end the inset at the end of the axis
+ */
+data class AxialInset(
```
turansky, line 36: `AxialInset` - is it name from specification?

turansky (changes requested):

kotlin-css/src/commonMain/kotlin/kotlinx/css/Inset.kt
CC007: No, but the alternative would have been to make 2 classes that have the exact same implementation: InsetInline and InsetBlock.

What do you think? Should I make AxialInset an abstract class with InsetLine and InsetBlock extending from AxialInset?

CC007: Hmm, the abstract class AxialInset doesn't add much beyond 2 one-liners, so this could be the alternative:

```kotlin

// ...

data class InsetInline(
    val start: LinearDimension,
    val end: LinearDimension,
) : CssValue(
    getShorthandValue(start, end)
) {
    override fun toString() = value

    constructor(all: LinearDimension) : this(all, all)
}

data class InsetBlock(
    val start: LinearDimension,
    val end: LinearDimension,
) : CssValue(
    getShorthandValue(start, end)
) {
    override fun toString() = value

    constructor(all: LinearDimension) : this(all, all)
}

@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.inset: Inset by CssProperty()
@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetInline: InsetInline by CssProperty()
@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetBlock: InsetBlock by CssProperty()

// ...
```

CC007: Or with typealias:

```kotlin
// ...

data class AxialInset(
    val start: LinearDimension,
    val end: LinearDimension,
) : CssValue(
    getShorthandValue(start, end)
) {
    override fun toString() = value

    constructor(all: LinearDimension) : this(all, all)
}

typealias InsetInline = AxialInset
typealias InsetBlock = AxialInset

@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.inset: Inset by CssProperty()
@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetInline: InsetInline by CssProperty()
@get:Deprecated("Write-only property", level = DeprecationLevel.HIDDEN) var StyledElement.insetBlock: InsetBlock by CssProperty()

// ...
```

CC007: Which solution would you prefer?

turansky: @CC007 sorry for delay.
I expect common type (like in `kotlin-cssom-core`).
Single (open) question - name.
In `kotlin-cssom-core` we names from specifications.
Here it should be something aligned with existed type names.

CC007: From what I can see in `kotlin-cssom-core`, all margin, scroll padding, text underline offset, inset and top/left/bottom/right types are type aliases of `AutoLengthProperty`, which is an interface that doesn't define its shape. It supports 1-value, 2-value, 3-value and 4-value properties. In fact, it doesn't seem to define anything about how many properties it contains. 

Margin and Inset do have helper functions to define the shape for those 4-value properties, but not for the 2-value properties like `InsetBlock` and `MarginInline`.

Introducing such a general type is I think out of scope for this PR and also doesn't add to the clarity of the parameters (start and end are specifically used here, just like in the spec, to avoid confusion when `direction: rtl` or `writing-mode: vertical-rl` is also used). 

If you don't want to introduce a new name, then I think that in this case the option with data class `InsetInline` and `InsetBlock` is best, since it doesn't require naming a superclass or typealias source.

CC007: Changed

turansky: > In fact, it doesn't seem to define anything about how many properties it contains.

You can use `array` method to provide 2 values ;)

turansky:
@CC007 thank you!

Current solution is fine for now.
Common type probably exist ([formal syntaxes are similar](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block#formal_syntax)), but it wasn't found in specifications.

